PR #6, opened 2016-07-28
Add efs impl that is able to be run as a standalone FileSystem

ellisjoe:
This change is [Reviewable](https://reviewable.io/reviews/palantir/hadoop-crypto/6)

### DIFF
--- a/hadoop-crypto/build.gradle
+++ b/hadoop-crypto/build.gradle
@@ -1,4 +1,5 @@
 apply from: "${rootDir}/gradle/publish.gradle"
+apply plugin: 'com.github.johnrengelman.shadow'
 
 sourceCompatibility = '1.7'
 
@@ -11,3 +12,39 @@ dependencies {
     testCompile "org.hamcrest:hamcrest-all"
     testCompile "org.mockito:mockito-core"
 }
+
+shadowJar {
+    mergeServiceFiles()
+
+    dependencies {
+        // only keep hadoop-aws, exclude all other hadoop-* jars.
+        exclude(dependency('org.apache.hadoop:hadoop-annotations'))
+        exclude(dependency('org.apache.hadoop:hadoop-auth'))
+        exclude(dependency('org.apache.hadoop:hadoop-client'))
+        exclude(dependency('org.apache.hadoop:hadoop-common'))
+        exclude(dependency('org.apache.hadoop:hadoop-hdfs'))
+        exclude(dependency('org.apache.hadoop:hadoop-mapreduce-client-app'))
+        exclude(dependency('org.apache.hadoop:hadoop-mapreduce-client-common'))
+        exclude(dependency('org.apache.hadoop:hadoop-mapreduce-client-core'))
+        exclude(dependency('org.apache.hadoop:hadoop-mapreduce-client-jobclient'))
+        exclude(dependency('org.apache.hadoop:hadoop-mapreduce-client-shuffle'))
+        exclude(dependency('org.apache.hadoop:hadoop-yarn-api'))
+        exclude(dependency('org.apache.hadoop:hadoop-yarn-client'))
+        exclude(dependency('org.apache.hadoop:hadoop-yarn-common'))
+        exclude(dependency('org.apache.hadoop:hadoop-yarn-server-common'))
+        exclude(dependency('org.apache.hadoop:hadoop-yarn-server-nodemanager'))
+    }
+
+    // Shade dependencies that can cause conflict with Hadoop
+    relocate('com.google.common', 'hadoop.crypto.shaded.com.google.common')
+    relocate('com.fasterxml.jackson', 'hadoop.crypto.shaded.com.fasterxml.jackson')
+    relocate('org.apache.commons.codec', 'hadoop.crypto.shaded.org.apache.commons.codec')
+    relocate('org.apache.commons.lang3', 'hadoop.crypto.shaded.org.apache.commons.lang3')
+    relocate('org.apache.commons.logging', 'hadoop.crypto.shaded.org.apache.commons.logging')
+    relocate('org.slf4j', 'hadoop.crypto.shaded.org.slf4j')
+}
+
+publishing.publications.nebula {
+    artifact(shadowJar)
+}
+
--- a/hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.hadoop;
+
+import com.google.common.base.Function;
+import com.google.common.base.Preconditions;
+import java.io.IOException;
+import java.net.URI;
+import java.security.KeyPair;
+import javax.ws.rs.core.UriBuilder;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.FilterFileSystem;
+import org.apache.hadoop.fs.Path;
+
+/**
+ * A {@link FileSystem} that transparently encrypts and decrypts the streams of an underlying FileSystem and stores
+ * encapsulated/wrapped encryption keys alongside the encrypted data in the same FileSystem as the encrypted data (see
+ * {@link FileKeyStorageStrategy}). Unlike the {@link EncryptedFileSystem}, the ConfigurableEncryptedFileSystem is
+ * initialized using a {@link Configuration} and can be used as a standalone FileSystem (like {@code s3a://, hdfs://,
+ * etc.}).
+ * <p>
+ * This FileSystem requires the first character of the scheme of any URI or Path to be `e` and the remainder of the
+ * scheme to be that of a backing FileSystem that can be instantiated using {@link FileSystem#get}.
+ */
+public final class ConfigurableEncryptedFileSystem extends FilterFileSystem {
+
+    /**
+     * Scheme is defined at runtime as `e[FS-scheme]`. This FileSystem can never be instantiated without
+     * `fs.e[FS-scheme].impl` being defined in the {@link Configuration} passed into the {@link #initialize} method.
+     */
+    private static final String SCHEME = "";
+    private static final String DEFAULT_ALGORITHM = "RSA";
+
+    /**
+     * Key mapping to a base64 encoded X509 public key.
+     */
+    public static final String PUBLIC_KEY_CONF = "fs.efs.key.public";
+
+    /**
+     * Key mapping to a base64 encoded PKCS8 private key.
+     */
+    public static final String PRIVATE_KEY_CONF = "fs.efs.key.private";
+
+    /**
+     * Key mapping to the public/private key algorithm (ex: RSA).
+     */
+    public static final String KEY_ALGORITHM_CONF = "fs.efs.key.algorithm";
+
+    private String encryptedScheme;
+
+    @Override
+    public void initialize(URI uri, Configuration conf) throws IOException {
+        encryptedScheme = uri.getScheme();
+        Preconditions.checkArgument(encryptedScheme.startsWith("e"),
+                "URI scheme must begin with 'e' but received: %s", encryptedScheme);
+
+        FileSystem delegate = getDelegateFileSystem(uri, conf);
+
+        KeyPair keyPair = getKeyPair(conf);
+        KeyStorageStrategy keyStore = new FileKeyStorageStrategy(delegate, keyPair);
+
+        this.fs = new EncryptedFileSystem(delegate, keyStore);
+    }
+
+    @Override
+    public URI getUri() {
+        return setUriSchemeFunc(encryptedScheme).apply(fs.getUri());
+    }
+
+    private static KeyPair getKeyPair(Configuration conf) {
+        String publicKey = Preconditions.checkNotNull(conf.get(PUBLIC_KEY_CONF),
+                "Public Key must be configured for key %s", PUBLIC_KEY_CONF);
+        String privateKey = conf.get(PRIVATE_KEY_CONF);
+        String algorithm = conf.get(KEY_ALGORITHM_CONF, DEFAULT_ALGORITHM);
+
+        return KeyPairs.fromStrings(privateKey, publicKey, algorithm);
+    }
+
+    private FileSystem getDelegateFileSystem(URI uri, Configuration conf) throws IOException {
+        String backingScheme = encryptedScheme.substring(1);
+        URI backingUri = setUriSchemeFunc(backingScheme).apply(uri);
+
+        FileSystem backingFs = FileSystem.get(backingUri, conf);
+        backingFs.initialize(backingUri, conf);
+
+        return new PathConvertingFileSystem(backingFs, setSchemeFunc(backingScheme), setSchemeFunc(encryptedScheme));
+    }
+
+    private static Function<Path, Path> setSchemeFunc(final String scheme) {
+        return new Function<Path, Path>() {
+            @Override
+            public Path apply(Path path) {
+                return new Path(setUriSchemeFunc(scheme).apply(path.toUri()));
+            }
+        };
+    }
+
+    private static Function<URI, URI> setUriSchemeFunc(final String scheme) {
+        return new Function<URI, URI>() {
+            @Override
+            public URI apply(URI uri) {
+                UriBuilder builder = UriBuilder.fromUri(uri);
+                if (uri.getScheme() != null) {
+                    builder.scheme(scheme);
+                }
+                return builder.build();
+            }
+        };
+    }
+
+    @Override
+    public String getScheme() {
+        return SCHEME;
+    }
+
+}
--- a/hadoop-crypto/src/main/java/com/palantir/hadoop/KeyPairs.java
+++ b/hadoop-crypto/src/main/java/com/palantir/hadoop/KeyPairs.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.hadoop;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Throwables;
+import java.security.GeneralSecurityException;
+import java.security.KeyFactory;
+import java.security.KeyPair;
+import java.security.PrivateKey;
+import java.security.spec.KeySpec;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
+import javax.annotation.Nullable;
+import org.apache.commons.net.util.Base64;
+
+public final class KeyPairs {
+
+    private KeyPairs() {}
+
+    /**
+     * Constructs a {@link KeyPair} from base64 encoded public and private keys.
+     *
+     * @param privateKeyString base64 encoded PKCS8 private key
+     * @param publicKeyString base64 encoded X509 public key
+     */
+    public static KeyPair fromStrings(@Nullable String privateKeyString, String publicKeyString, String algorithm) {
+        try {
+            KeyFactory keyFactory = KeyFactory.getInstance(algorithm);
+            // Private key is only required for decryption, can be null
+            Optional<PrivateKey> privateKey = Optional.absent();
+            if (privateKeyString != null) {
+                KeySpec privateKs = new PKCS8EncodedKeySpec(Base64.decodeBase64(privateKeyString));
+                privateKey = Optional.of(keyFactory.generatePrivate(privateKs));
+            }
+
+            KeySpec publicKs = new X509EncodedKeySpec(Base64.decodeBase64(publicKeyString));
+
+            return new KeyPair(keyFactory.generatePublic(publicKs), privateKey.orNull());
+        } catch (GeneralSecurityException e) {
+            throw Throwables.propagate(e);
+        }
+    }
+
+}
--- a/hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
+++ b/hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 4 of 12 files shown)


ellisjoe:
@uschi2000 still need to add fat jar publishing and docs. impls all there though.

ellisjoe:
@markelliot this is the piece that lets you write as a standalone operation.

.baseline/checkstyle/checkstyle.xml
```diff
@@ -307,7 +307,7 @@
         </module>
         <module name="RegexpSinglelineJava">
             <property name="format" value="\bIOUtils\.toString\("/>
-            <property name="message" value="Prefer Guava''s {CharStreams,Files,Resources}.toString to avoid charset/stream closing issues."/>
+            <property name="message" value="Prefer Guava''s [CharStreams,Files,Resources].toString to avoid charset/stream closing issues."/>
```
ellisjoe, line 296: @uschi2000 needed to change this otherwise checkstyle throws:

```
Caused by: java.lang.IllegalArgumentException: can't parse argument number: CharStreams
        at com.puppycrawl.tools.checkstyle.api.LocalizedMessage.getMessage(LocalizedMessage.java:265)
        at com.puppycrawl.tools.checkstyle.api.LocalizedMessage.compareTo(LocalizedMessage.java:377)
        at com.puppycrawl.tools.checkstyle.api.LocalizedMessage.compareTo(LocalizedMessage.java:49)
        at com.puppycrawl.tools.checkstyle.api.LocalizedMessages.add(LocalizedMessages.java:53)
        at com.puppycrawl.tools.checkstyle.api.Check.log(Check.java:239)
        at com.puppycrawl.tools.checkstyle.checks.regexp.SinglelineDetector.checkLine(SinglelineDetector.java:114)
        at com.puppycrawl.tools.checkstyle.checks.regexp.SinglelineDetector.processLines(SinglelineDetector.java:52)
        at com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineJavaCheck.beginTree(RegexpSinglelineJavaCheck.java:73)
        at com.puppycrawl.tools.checkstyle.TreeWalker.notifyBegin(TreeWalker.java:354)
        at com.puppycrawl.tools.checkstyle.TreeWalker.walk(TreeWalker.java:326)
        at com.puppycrawl.tools.checkstyle.TreeWalker.processFiltered(TreeWalker.java:203)
        at com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.process(AbstractFileSetCheck.java:79)
        at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:265)
        at com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask.processFiles(CheckstyleAntTask.java:351)
        at com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask.realExecute(CheckstyleAntTask.java:320)
        at com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask.execute(CheckstyleAntTask.java:292)
        at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
        at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:106)
        ... 88 more
Caused by: java.lang.NumberFormatException: For input string: "CharStreams"
        ... 106 more
```

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -0,0 +1,80 @@
...
+        String privateKey = conf.get(PRIVATE_KEY_CONF);
+        String algorithm = conf.get(KEY_ALGORITHM_CONF, DEFAULT_ALGORITHM);
+
+        Preconditions.checkNotNull(publicKey, "Public Key must be configured for key %s", PUBLIC_KEY_CONF);
+        return KeyPairs.fromStrings(privateKey, publicKey, algorithm);
+    }
+
+    private static FileSystem getBackingFileSystem(URI uri, Configuration conf) throws IOException {
+        Class<? extends FileSystem> fileSystemClass =
+                conf.getClass(BACKING_FILESYSTEM_CONF, RawLocalFileSystem.class, FileSystem.class);
```
ellisjoe: defaulting to `RawLocalFileSystem` for now since that makes it easy to test this out without an s3 bucket or hdfs running. open to other options though

.baseline/checkstyle/checkstyle.xml
uschi2000, line 296: already changed upstream: https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml#L302

uschi2000, line 296: xpy734

uschi2000, line 296: (that's the code to unlock the hobbitworld, according to Caitlin's sister.)

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -0,0 +1,80 @@
...
+import java.net.URI;
+import java.security.KeyPair;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.FilterFileSystem;
+import org.apache.hadoop.fs.RawLocalFileSystem;
+import org.apache.hadoop.util.ReflectionUtils;
+
+/**
+ * A {@link FileSystem} that encrypts and decrypts the streams of an underlying FileSystem and stores envelope
```
uschi2000: the thing is called key encapsulation: https://en.wikipedia.org/wiki/Key_encapsulation

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -0,0 +1,80 @@
...
+ * encryption keys on the same FileSystem as the encrypted data (see {@link FileKeyStorageStrategy}). Unlike the {@link
```
uschi2000: ... and stores encapsulated/wrapped encryption alongside the encrypted data in the same FileSystem

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -0,0 +1,80 @@
...
+ * EncryptedFileSystem}, the ConfigurableEncryptedFileSystem is initialized using a {@link Configuration} and can be
+ * used as a standalone FileSystem.
```
uschi2000: ... standalone Hadoop FileSystem like `s3://`, `s3a://`, or `hdfs://`.

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
uschi2000: ... that transparently encrypts and decrypts ...

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -0,0 +1,80 @@
...
+
+        KeyPair keyPair = getKeyPair(conf);
+        KeyStorageStrategy keyStore = new FileKeyStorageStrategy(backingFileSystem, keyPair);
+
+        this.fs = new EncryptedFileSystem(backingFileSystem, keyStore);
+        super.initialize(uri, conf);
+    }
+
+    private static KeyPair getKeyPair(Configuration conf) {
+        String publicKey = conf.get(PUBLIC_KEY_CONF);
```
uschi2000: how about?

```
String publicKey = Preconditons.checkNotNull(publicKey, "...");
```

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
uschi2000: I think you should error here if it's not set.

hadoop-crypto/src/main/java/com/palantir/hadoop/KeyPairs.java
```diff
@@ -0,0 +1,53 @@
...
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
+import javax.annotation.Nullable;
+import org.apache.commons.net.util.Base64;
+
+public final class KeyPairs {
+
+    private KeyPairs() {}
+
+    public static KeyPair fromStrings(@Nullable String privateKeyString, String publicKeyString, String algorithm) {
```
uschi2000, line 41: needs some docs somewhere on the allowed values for these strings

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.hadoop.fs.RawLocalFileSystem;
+import org.apache.hadoop.util.ReflectionUtils;
+
+/**
+ * A {@link FileSystem} that encrypts and decrypts the streams of an underlying FileSystem and stores envelope
+ * encryption keys on the same FileSystem as the encrypted data (see {@link FileKeyStorageStrategy}). Unlike the {@link
+ * EncryptedFileSystem}, the ConfigurableEncryptedFileSystem is initialized using a {@link Configuration} and can be
+ * used as a standalone FileSystem.
+ */
+public final class ConfigurableEncryptedFileSystem extends FilterFileSystem {
```
uschi2000, line 40: is there a way we can name this EncryptedFileSystem without clashes?

hadoop-crypto/src/main/java/com/palantir/hadoop/KeyPairs.java
uschi2000, line 41: can this be protected?

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
```diff
@@ -0,0 +1,141 @@
...
+    @Before
+    public void before() throws IOException {
+        KeyPair keyPair = TestKeyPairs.generateKeyPair();
+        Configuration conf = getBaseConf();
+        conf.set(ConfigurableEncryptedFileSystem.PUBLIC_KEY_CONF,
+                Base64.encodeBase64String(keyPair.getPublic().getEncoded()));
+        conf.set(ConfigurableEncryptedFileSystem.BACKING_FILESYSTEM_CONF,
+                RawLocalFileSystem.class.getCanonicalName());
+
+        efsPublic = FileSystem.newInstance(EFS_URI, conf);
```
uschi2000: only used in one test: move there?

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+        // Write encrypted data
+        OutputStream os = efs.create(path);
+        IOUtils.write(data, os);
+        os.close();
+
+        // Read encrypted data
+        InputStream dis = efs.open(path);
+        IOUtils.readFully(dis, readData);
+        assertThat(readData, is(dataBytes));
```
uschi2000, line 89: I assume that `is()` compares array content? correct?

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+        // Raw data is not the same
+        dis = rawFs.open(path);
+        IOUtils.readFully(dis, readData);
+        assertThat(readData, is(not(dataBytes)));
+
+        // KeyMaterial file exists
+        assertTrue(efs.exists(new Path(path + FileKeyStorageStrategy.EXTENSION)));
```
uschi2000: maybe use rawFs here for clarity?

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Test
+    public void testOnlyPublicKey() throws IOException {
+        String data = "data";
+
+        // Write encrypted data
+        OutputStream os = efsPublic.create(path);
+        IOUtils.write(data, os);
+        os.close();
```
uschi2000, line 112: assert that data & key are in rawFs

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
```diff
@@ -0,0 +1,141 @@
...
+    @Test
+    public void testBackingFsInvalid() throws IOException {
+        Configuration conf = getBaseConf();
+        conf.set(ConfigurableEncryptedFileSystem.BACKING_FILESYSTEM_CONF, String.class.getCanonicalName());
+
+        expectedException.expect(RuntimeException.class);
+        expectedException.expectMessage(
+                String.format("%s not %s", String.class.getCanonicalName(), FileSystem.class.getCanonicalName()));
+        FileSystem.newInstance(EFS_URI, conf);
+    }
```
uschi2000, line 168: do we have a test that exemplifies what happens when file names clash?

```
efs.create("foo");
efs.create("foo" + FileKeyStorageStrategy.EXTENSION);
efs.open("foo"); // error, key is invalid
```

uschi2000:
few nitpicks, but looks cool in general!

uschi2000:
should add to README how to use this thing

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe, line 40: move it to a different package but their both going to end up being public. I'll think about other options

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/KeyPairs.java
ellisjoe, line 41: added docs. We'd like to use this in other projects so would prefer public.

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
ellisjoe: moved. unsetting the private key from the conf in the test now

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
ellisjoe, line 89: yeah. `equalTo` check to see if the type is an array and then does an array comparison. the test below this correctly checks that the arrays are NOT equal so that gives me confidence its doing the right thing

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
ellisjoe: done

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
ellisjoe, line 112: done

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
ellisjoe, line 168: we do now

ellisjoe:
still need docs and shadowJar publishing. also consider renaming `ConfigurableEncryptedFileSystem`

uschi2000:
This thing doesn't work, correct, Joe?

ellisjoe:
correct. still working on a reasonable solution

hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
```diff
@@ -49,7 +49,7 @@ public PathConvertingFileSystem(FileSystem delegate,
 
     @Override
     public URI getUri() {
-        return delegate.getUri();
+        return from(new Path(delegate.getUri())).toUri();
```
ellisjoe: this is slightly disgusting hack

ellisjoe:
@uschi2000 good to review. currently depends on the rename PR to get this working with the cli for puts. ended up going the route of declaring the backing fs as part of the scheme (ie: `efile` or `es3a`) rather than sticking with `efs` always and specifying the fs in config.

hadoop-crypto/build.gradle
```diff
@@ -11,3 +12,39 @@ dependencies {
...
+        exclude(dependency('org.apache.hadoop:hadoop-yarn-server-nodemanager'))
+    }
+
+    // Shade dependencies that can cause conflict with Hadoop
+    relocate('com.google.common', 'hadoop.crypto.shaded.com.google.common')
+    relocate('com.fasterxml.jackson', 'hadoop.crypto.shaded.com.fasterxml.jackson')
+    relocate('org.apache.commons.codec', 'hadoop.crypto.shaded.org.apache.commons.codec')
+    relocate('org.apache.commons.lang3', 'hadoop.crypto.shaded.org.apache.commons.lang3')
+    relocate('org.apache.commons.logging', 'hadoop.crypto.shaded.org.apache.commons.logging')
+    relocate('org.slf4j', 'hadoop.crypto.shaded.org.slf4j')
```
uschi2000, line 44: are you sure relocation slf4j works?

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * A {@link FileSystem} that transparently encrypts and decrypts the streams of an underlying FileSystem and stores
+ * encapsulated/wrapped encryption keys alongside the encrypted data in the same FileSystem as the encrypted data (see
+ * {@link FileKeyStorageStrategy}). Unlike the {@link EncryptedFileSystem}, the ConfigurableEncryptedFileSystem is
+ * initialized using a {@link Configuration} and can be used as a standalone FileSystem (like {@code s3a://, hdfs://,
+ * etc.}).
+ */
+public final class ConfigurableEncryptedFileSystem extends FilterFileSystem {
+
+    private static final String SCHEME = "";
```
uschi2000, line 46: what does this do? probably justifies a comment

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -0,0 +1,106 @@
...
+    private static final String DEFAULT_ALGORITHM = "RSA";
+
+    public static final String PUBLIC_KEY_CONF = "fs.efs.key.public";
```
uschi2000, line 52: what format is expected for these? should also say that these are keys into the given Configuration object

hadoop-crypto/src/main/java/com/palantir/hadoop/KeyPairs.java
```diff
@@ -0,0 +1,59 @@
...
+import java.security.spec.X509EncodedKeySpec;
+import javax.annotation.Nullable;
+import org.apache.commons.net.util.Base64;
+
+public final class KeyPairs {
+
+    private KeyPairs() {}
+
+    /**
+     * Construct a {@link KeyPair} from base64 encoded public and private keys.
```
uschi2000: Construct_S_

hadoop-crypto/src/main/java/com/palantir/hadoop/KeyPairs.java
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param privateKeyString base64 encoded PKCS8 private key
```
uschi2000, line 38: inline with above?

hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
uschi2000: why is this required? (do you have test that fails if you used `delegate.getUri()`?)

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
```diff
@@ -0,0 +1,171 @@
...
+        os.close();
+
+        // Write encrypted data
+        os = efs.create(new Path(path + FileKeyStorageStrategy.EXTENSION));
+        IOUtils.write(data, os);
+        os.close();
+
+        // NOTE(jellis): This is the most likely exception, however if the first byte of the .keymaterial file is the
+        // same as the KeyMaterials version then any number of RuntimeExceptions may be thrown.
+        // expectedException.expect(IllegalArgumentException.class);
```
uschi2000, line 162: huh?

uschi2000:
looks good modulo a few nitpicks

uschi2000 (dismissed):

hadoop-crypto/build.gradle
ellisjoe: this is what we do elsewhere. anything in particular you're concerned with?

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/KeyPairs.java
ellisjoe: tried it but think this a bit nicer (looks pretty cluttered when I inlined it)

hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
ellisjoe: fixed this. now the ConfigurableEFS properly sets the scheme in getUri

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
ellisjoe: the goal of the test is to say if someone writes a file ending in `.keymaterial` then the EFS fails to decrypt the file whos keymaterial has been overwritten. you'd expect it to fail due to incorrect format version most of the time but it's possible for the first encrypted byte of the new file to be 1 or 2 which would cause the EFS to fail for any number of reasons. may not even be worth this comment though.

uschi2000:
+1

PR #9, opened 2016-09-10
Jellis/path converting fs

ellisjoe:
@uschi2000 @markelliot need this for the `efs` -> `{hdfs,s3a,...}` scheme conversions.

Note: This only implements the required `FileSystem` methods and not every method, which means some of the methods implemented by the underlying `FileSystem` won't work.

### DIFF
--- a/hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
@@ -109,12 +109,10 @@ public final class PathConvertingFileSystem extends FileSystem {
         return delegate.mkdirs(toDelegatePath(f), permission);
     }
 
-    // convenience
     private Path toDelegatePath(Path path) {
         return toDelegatePathFunc.apply(path);
     }
 
-    // convenience
     private Path toReturnPath(Path path) {
         return toReturnPathFunc.apply(path);
     }
--- a/hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
@@ -29,8 +29,9 @@ import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.util.Progressable;
 
 /**
- * A decorator {@link FileSystem} that converts input paths using {@code toDelegateUri} and returned paths to {@code
- * fromScheme}.
+ * A decorator {@link FileSystem} that delegates calls and converts paths to/from the delegate FileSystem.
+ * {@link Path}s passed into methods are converted using {@code toDelegatePathFunc} before being forwarded.
+ * {@link Path}s returned from the delegate are converted using {@code toReturnPathFunc} before being returned.
  */
 public final class PathConvertingFileSystem extends FileSystem {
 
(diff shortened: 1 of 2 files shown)


hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * A decorator {@link FileSystem} that delegates calls and converts paths to/from the delegate FileSystem.
+ * {@link Path}s passed into methods are converted using {@code toDelegatePathFunc} before being forwarded.
+ * {@link Path}s returned from the delegate are converted using {@code toReturnPathFunc} before being returned.
+ */
+public final class PathConvertingFileSystem extends FileSystem {
+
+    private final FileSystem delegate;
+    private final Function<Path, Path> toDelegatePathFunc;
```
uschi2000: `toDelegatePath`, `fromDelegatePath`

uschi2000: or maybe even `to`, `from`

hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.permission.FsPermission;
+import org.apache.hadoop.util.Progressable;
+
+/**
+ * A decorator {@link FileSystem} that delegates calls and converts paths to/from the delegate FileSystem.
+ * {@link Path}s passed into methods are converted using {@code toDelegatePathFunc} before being forwarded.
```
uschi2000: `@link #to...`

hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
```diff
@@ -0,0 +1,136 @@
...
+ * {@link Path}s returned from the delegate are converted using {@code toReturnPathFunc} before being returned.
```
uschi2000: `@link #from...`

hadoop-crypto/src/test/java/com/palantir/hadoop/PathConvertingFileSystemTest.java
```diff
@@ -0,0 +1,165 @@
...
+        boolean success = convertingFs.mkdirs(PATH, null);
+
+        assertThat(success, is(false));
+    }
+
+    private static FileStatus fileStatus(Path path) {
+        return new FileStatus(0, false, 0, 0, 0, path);
+    }
+
+    private enum DelegatePathFunc implements Function<Path, Path> {
```
uschi2000: move these two to the top

hadoop-crypto/src/test/java/com/palantir/hadoop/PathConvertingFileSystemTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        assertThat(actualUri, is(DUMMY_URI));
+    }
+
+    @Test
+    public void open() throws Exception {
+        when(delegate.open(DELEGATE_PATH, 0)).thenReturn(inputStream);
+        FSDataInputStream actualStream = convertingFs.open(PATH, 0);
+
+        assertThat(actualStream, is(inputStream));
```
uschi2000, line 88: why don't you use `verify()` to assert that the delegate is called with the expected (converted) parameters?

ellisjoe: started that way but realized I also want to verify that the return value is correct since it's possible to call the delegate correctly without propagating what the delegate returns.

hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/PathConvertingFileSystem.java
ellisjoe: changed the function parameters to `toFunc` and `fromFunc` and then my helper methods to `to` and `from`.

hadoop-crypto/src/test/java/com/palantir/hadoop/PathConvertingFileSystemTest.java
ellisjoe: done

uschi2000:
:+1:

PR #10, opened 2016-09-27
Fix KeyMaterial serialization

ellisjoe:
- the length of several byte[] were being written as single bytes rather than
  integers which meant byte arrays longer than 255 bytes would have their size incorrectly serialized. When using 2048 bit RSA keys with 256 bit AES keys the resulting wrapped key is of length 256 which meant its length would be incorrectly serialized and deserialization would fail.

### DIFF
--- a/hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeyMaterials.java
+++ b/hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeyMaterials.java
@@ -41,7 +41,7 @@ import org.slf4j.LoggerFactory;
 public final class KeyMaterials {
 
     private static final Logger log = LoggerFactory.getLogger(KeyMaterials.class);
-    private static final Map<Integer, KeySerializer> SERIALIZERS = KeySerializers.getSerializers();
+    private static final Map<Integer, ? extends KeySerializer> SERIALIZERS = KeySerializers.getSerializers();
 
     private KeyMaterials() {}
 
@@ -81,7 +81,7 @@ public final class KeyMaterials {
     }
 
     public static byte[] wrap(KeyMaterial keyMaterial, PublicKey key) {
-        return SERIALIZERS.get(KeySerializerV2.VERSION).wrap(keyMaterial, key);
+        return KeySerializerV2.INSTANCE.wrap(keyMaterial, key);
     }
 
     public static KeyMaterial unwrap(byte[] wrappedKeyMaterial, PrivateKey key) {
--- a/hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeyMaterials.java
+++ b/hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeyMaterials.java
@@ -18,10 +18,9 @@ package com.palantir.hadoop;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableMap;
 import com.palantir.hadoop.serializer.KeySerializer;
-import com.palantir.hadoop.serializer.KeySerializerV1;
 import com.palantir.hadoop.serializer.KeySerializerV2;
+import com.palantir.hadoop.serializer.KeySerializers;
 import java.io.ByteArrayInputStream;
 import java.io.DataInputStream;
 import java.io.IOException;
@@ -45,9 +44,7 @@ import org.slf4j.LoggerFactory;
 public final class KeyMaterials {
 
     private static final Logger log = LoggerFactory.getLogger(KeyMaterials.class);
-    private static final Map<Integer, KeySerializer> SERIALIZAERS = ImmutableMap.of(
-            KeySerializerV1.VERSION, new KeySerializerV1(),
-            KeySerializerV2.VERSION, new KeySerializerV2());
+    private static final Map<Integer, KeySerializer> SERIALIZERS = KeySerializers.getSerializers();
 
     private KeyMaterials() {}
 
@@ -87,7 +84,7 @@ public final class KeyMaterials {
     }
 
     public static byte[] wrap(KeyMaterial keyMaterial, PublicKey key) {
-        return SERIALIZAERS.get(KeySerializerV2.VERSION).wrap(keyMaterial, key);
+        return SERIALIZERS.get(KeySerializerV2.VERSION).wrap(keyMaterial, key);
     }
 
     public static KeyMaterial unwrap(byte[] wrappedKeyMaterial, PrivateKey key) {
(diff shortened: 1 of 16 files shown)


ellisjoe:
@uschi2000 @markelliot most of the code change is a refactor to pull out the serialization code into KeySerializerV1. KeySerializerV2 is a copy of V1 except that calls to `read()` and `write()` are now replaced by `readInt()` and `writeInt()` to fix the problem stated above. The rest is test code.

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeySerializer.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    byte[] wrap(KeyMaterial keyMaterial, PublicKey key);
+
+    /**
+     * Deserializes and unwraps a {@link KeyMaterial} from the provided {@code byte[]}. The input {@code byte[]} is only
+     * guaranteed to be valid if produced by the corresponding {@link #wrap} method.
+     */
+    KeyMaterial unwrap(byte[] wrappedKeyMaterial, PrivateKey key);
+
+    int getVersion();
```
uschi2000, line 39: comment

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeySerializerV1.java
```diff
@@ -0,0 +1,114 @@
...
+ * serialized as follows, which is the same format the {@link #unwrap} method expects:
+ *
+ * <pre>
+ *  +--------------------------------------------------------------------------------------------------------------+
+ *  | version | cipher algorithm length | cipher algorithm | wrapped key length | wrapped key | iv length |   iv   |
+ *  |   byte  |           byte          |       byte[]     |         byte       |    byte[]   |    byte   | byte[] |
+ *  +--------------------------------------------------------------------------------------------------------------+
+ * </pre>
+ */
+final class KeySerializerV1 implements KeySerializer {
```
uschi2000: `@Deprecated` ?

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeySerializerV1.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Serializer of wrapping and unwrapping {@link KeyMaterial}. The {@link #wrap} method returns the KeyMaterial
+ * serialized as follows, which is the same format the {@link #unwrap} method expects:
+ *
+ * <pre>
+ *  +--------------------------------------------------------------------------------------------------------------+
+ *  | version | cipher algorithm length | cipher algorithm | wrapped key length | wrapped key | iv length |   iv   |
+ *  |   byte  |           byte          |       byte[]     |         byte       |    byte[]   |    byte   | byte[] |
+ *  +--------------------------------------------------------------------------------------------------------------+
+ * </pre>
```
uschi2000, line 43: should add warning here indicating that this thing does not work if `algorithm`, `key`, or `iv` is longer than 255 bytes

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeyMaterials.java
```diff
@@ -98,79 +81,33 @@ public static KeyMaterial deserialize(SerializableKeyMaterial keyMaterial) {
...
-            byte[] iv = keyMaterial.getIv();
-            stream.write(iv.length);
-            stream.write(iv);
-
-            stream.close();
-            return stream.toByteArray();
-        } catch (IOException | InvalidKeyException | IllegalBlockSizeException e) {
-            throw Throwables.propagate(e);
-        }
+        return SERIALIZERS.get(KeySerializerV2.VERSION).wrap(keyMaterial, key);
```
uschi2000: no need for lookup

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeySerializerV2.java
```diff
@@ -0,0 +1,121 @@
...
+import javax.crypto.SecretKey;
+
+/**
+ * Serializer of wrapping and unwrapping {@link KeyMaterial}. The {@link #wrap} method returns the KeyMaterial
+ * serialized as follows, which is the same format the {@link #unwrap} method expects:
+ *
+ * <pre>
+ *  +--------------------------------------------------------------------------------------------------------------+
+ *  | version | cipher algorithm length | cipher algorithm | wrapped key length | wrapped key | iv length |   iv   |
+ *  |   byte  |           int           |       byte[]     |         int        |    byte[]   |    int    | byte[] |
```
uschi2000, line 43: only 253 bugs left in this format.

hadoop-crypto/src/test/java/com/palantir/hadoop/KeyMaterialsTest.java
```diff
@@ -64,15 +64,30 @@ public void testWrapAndUnwrap() {
     }
```
uschi2000, line 74: add test verifying that default serializer is v2

uschi2000 (commented):
looks good modulo few nitpicks

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeyMaterials.java
ellisjoe: needed since the methods are no longer static.

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeySerializer.java
ellisjoe: done

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeySerializerV1.java
ellisjoe: done

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeySerializerV1.java
ellisjoe: done

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeySerializerV2.java
ellisjoe: at the current rate we've got about 200 years until we run out :)

If we ever get past version 100 we could switch to interpreting is as a UTF-8 encoded value

hadoop-crypto/src/test/java/com/palantir/hadoop/KeyMaterialsTest.java
ellisjoe: done

hadoop-crypto-api/src/main/java/com/palantir/hadoop/KeyMaterials.java
uschi2000: i don't follow; you always want a KeySerializerV2 here, why do you have to grab it from a Map<>?

ellisjoe: I need `KeySerializerV2#wrap(...)` so I can either grab it from the map or instantiate a new one. I guess I could also just have a static one in KeyMaterials? was just avoiding creating new ones (thinking in terms of something like `ObjectMappers.getMapper()`)

uschi2000: the serializers are stateless, right? proposal:
- make them enum singletons
- put the singletons in the map
- use the v2 singleton directly here

ellisjoe: done

uschi2000:
👍

uschi2000:
next up: blog post.

uschi2000 (approved):

PR #12, opened 2016-10-13
Implement rename in EncryptedFileSystem

ellisjoe:


ellisjoe:
@uschi2000 @markelliot needed to get efs to work with the cli. will implement `delete` and `append` later

hadoop-crypto/src/main/java/com/palantir/hadoop/EncryptedFileSystem.java
```diff
@@ -76,6 +79,23 @@ public FSDataOutputStream create(Path path, FsPermission permission,
...
 
+    @Override
+    public boolean rename(Path src, Path dst) throws IOException {
+        KeyMaterial keyMaterial = keyStore.get(src.toString());
+        keyStore.put(dst.toString(), keyMaterial);
+        boolean renamed = fs.rename(src, dst);
+
+        if (renamed) {
+            try {
+                keyStore.remove(src.toString());
```
markelliot: Don't you need to put the key material back in the new place?

ellisjoe: it's put into the new place before renaming the file

ellisjoe: added `keyStore.remove(dst)` when the rename fails

hadoop-crypto/src/main/java/com/palantir/hadoop/EncryptedFileSystem.java
```diff
@@ -76,6 +79,29 @@ public FSDataOutputStream create(Path path, FsPermission permission,
         return os;
     }
 
+    @Override
+    public boolean rename(Path src, Path dst) throws IOException {
```
uschi2000, line 83: is the path guaranteed to be sanitized when it gets here? for instance, what happens if I `rename("foo", "bar//baz")` and then `get("bar/baz")` ?

hadoop-crypto/src/main/java/com/palantir/hadoop/EncryptedFileSystem.java
```diff
@@ -76,6 +79,29 @@ public FSDataOutputStream create(Path path, FsPermission permission,
...
+        }
+
+        return renamed;
+    }
+
+    private void tryRemoveKey(Path path) {
+        try {
+            keyStore.remove(path.toString());
+        } catch (Exception e) {
+            log.warn("Unable to remove KeyMaterial for file: {}", path);
```
uschi2000, line 103: attach e

hadoop-crypto/src/main/java/com/palantir/hadoop/EncryptedFileSystem.java
```diff
@@ -76,6 +79,29 @@ public FSDataOutputStream create(Path path, FsPermission permission,
         return os;
     }
 
+    @Override
+    public boolean rename(Path src, Path dst) throws IOException {
+        KeyMaterial keyMaterial = keyStore.get(src.toString());
+        keyStore.put(dst.toString(), keyMaterial);
```
uschi2000, line 87: leave comment explaining the convoluted code path (provides safety net in case `put` fails)

uschi2000 (changes requested):
looks good module a few comments

hadoop-crypto/src/main/java/com/palantir/hadoop/EncryptedFileSystem.java
ellisjoe: `Path.toString()` returns the canonical path string so `new Path("bar//baz").toString()` gives `"bar/baz"`. that's what we use to put/get keys so that should be fine. anything else is on the underlying filesystem

uschi2000: OK, grand. Might be worth adding a test because that behaviour seems somewhat brittle?

ellisjoe: same thing could be said for `create` and `open` right? maybe the test should just be call one of the FileSystem methods with a path like `foo//bar` and make sure the keystore gets called with `foo/bar`?

uschi2000:
+1

hadoop-crypto/src/main/java/com/palantir/hadoop/EncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/EncryptedFileSystem.java
ellisjoe: don't think that's right. `tryRemoveKey` is supposed to be best effort removal. The signal for failure to rename is the boolean returned by the method.

hadoop-crypto/src/main/java/com/palantir/hadoop/EncryptedFileSystem.java
ellisjoe: added tests `test{create,open,rename}_normalizePathPassedToKeyStore`

uschi2000 (approved):

PR #13, opened 2016-10-19
Jellis/efs docs

ellisjoe:


README.md
```diff
@@ -20,8 +20,48 @@ which stores the key for future access. The provided storage strategy
...
+```xml
+<configuration>
+    <property>
+        <name>fs.efile.impl</name> <!-- others: fs.es3a.impl or fs.ehdfs.impl -->
+        <value>com.palantir.hadoop.ConfigurableEncryptedFileSystem</value>
+    </property>
+
+    <property>
+        <name>fs.efs.key.public</name>
+        <value>REDACTED</value>
```
uschi2000: Should mention format. Maybe just generate a pair and leave a comment saying to generate them?

README.md
```diff
@@ -20,8 +20,48 @@ which stores the key for future access. The provided storage strategy
...
+    <property>
+        <name>fs.efs.key.private</name>
+        <value>REDACTED</value>
+    </property>
+</configuration>
+```
+
+##### Commands
+``` bash
+./bin/hadoop dfs -put file.txt /tmp/file.txt
```
uschi2000: Why would this use efile and not file? Maybe better to use paths with schema.

README.md
uschi2000: I see you mention the format below. Would still be good to give an example and show how to generate then in linux

uschi2000 (changes requested):

README.md
ellisjoe: whoops. fixed

README.md
ellisjoe: done

uschi2000:
+1

PR #14, opened 2016-10-20
filter .keymaterial files in listStatus

ellisjoe:
@uschi2000

### DIFF
--- a/hadoop-crypto/src/main/java/com/palantir/hadoop/StandaloneEncryptedFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/hadoop/StandaloneEncryptedFileSystem.java
@@ -18,11 +18,16 @@ package com.palantir.hadoop;
 
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
+import com.google.common.collect.Collections2;
 import java.io.IOException;
 import java.net.URI;
 import java.security.KeyPair;
+import java.util.Arrays;
+import java.util.Collection;
 import javax.ws.rs.core.UriBuilder;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.FilterFileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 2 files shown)


hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -82,6 +87,19 @@ public URI getUri() {
         return setUriSchemeFunc(encryptedScheme).apply(fs.getUri());
     }
 
+    @Override
+    // TODO(jellis): consider moving logic related to FileKeyStorageStrategy into a separate FileSystem
+    public FileStatus[] listStatus(Path path) throws IOException {
+        Collection<FileStatus> files = Collections2.filter(Arrays.asList(fs.listStatus(path)),
+                new Predicate<FileStatus>() {
```
uschi2000: static?

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
```diff
@@ -82,6 +87,19 @@ public URI getUri() {
         return setUriSchemeFunc(encryptedScheme).apply(fs.getUri());
     }
 
+    @Override
+    // TODO(jellis): consider moving logic related to FileKeyStorageStrategy into a separate FileSystem
+    public FileStatus[] listStatus(Path path) throws IOException {
+        Collection<FileStatus> files = Collections2.filter(Arrays.asList(fs.listStatus(path)),
```
uschi2000: hmm, should we avoid going array->list->array?

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
```diff
@@ -167,6 +169,20 @@ public void testFileNameCollidesWithKeyMaterial() throws IOException {
         efs.open(path);
     }
 
+    @Test
+    public void testListStatus_keyMaterialFilesFiltered() throws IOException {
```
uschi2000: curious why we never hit this bug before?

ellisjoe: a FileSystem higher up intercepted the listStatus calls and never directed them at the backing store

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/hadoop/ConfigurableEncryptedFileSystem.java
ellisjoe: Arrays.asList doesn't copy so we should be fine. Gonna need array -> array anyway so this should be equivalent.

hadoop-crypto/src/test/java/com/palantir/hadoop/ConfigurableEncryptedFileSystemTest.java
uschi2000: oh, i see

PR #17, opened 2016-11-04
Override InputStream#read(byte[], int, int) for performance

schlosna:
The following backtrace shows previously doing single byte read() which is inefficient, so lets support bulk reads.

```
com.sun.crypto.provider.CipherCore.update(byte[], int, int)
com.sun.crypto.provider.AESCipher.engineUpdate(byte[], int, int)
javax.crypto.Cipher.update(byte[], int, int)
javax.crypto.CipherInputStream.getMoreData()
javax.crypto.CipherInputStream.read()
com.palantir.hadoop.cipher.FsCipherInputStream.read()
java.io.InputStream.read(byte[], int, int)
```

### DIFF
--- a/hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
+++ b/hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
@@ -100,7 +100,6 @@ public final class FsCipherInputStream extends FSInputStream {
         }
     }
 
-    // NOTE(jellis): may want to implement the other read methods depending on performance
     @Override
     public int read() throws IOException {
         int bytesRead = decryptedStream.read();
--- a/hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
+++ b/hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
@@ -122,5 +122,6 @@ public final class FsCipherInputStream extends FSInputStream {
     @Override
     public void close() throws IOException {
         delegate.close();
+        decryptedStream.close();
     }
 }
--- a/hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
+++ b/hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
@@ -103,11 +103,20 @@ public final class FsCipherInputStream extends FSInputStream {
     // NOTE(jellis): may want to implement the other read methods depending on performance
     @Override
     public int read() throws IOException {
-        int ret = decryptedStream.read();
-        if (ret != -1) {
+        int bytesRead = decryptedStream.read();
+        if (bytesRead != -1) {
             decryptedStreamPos++;
         }
-        return ret;
+        return bytesRead;
+    }
+
+    @Override
+    public int read(byte[] buffer, int offset, int length) throws IOException {
+        int bytesRead = decryptedStream.read(buffer, offset, length);
+        if (bytesRead != -1) {
+            decryptedStreamPos += bytesRead;
+        }
+        return bytesRead;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


schlosna:
@ellisjoe @pwoody @robert3005 @markelliot for SA

uschi2000:
Yeah, I had suggested that a couple of months back, but I guess nobody every verified performance benefits. Do you have numbers or evidence, @schlosna ?

hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
```diff
@@ -103,15 +103,25 @@ public boolean seekToNewSource(long targetPos) throws IOException {
...
+        int bytesRead = decryptedStream.read();
+        if (bytesRead != -1) {
             decryptedStreamPos++;
         }
-        return ret;
+        return bytesRead;
+    }
+
+    @Override
+    public int read(byte[] buffer, int offset, int length) throws IOException {
```
uschi2000, line 113: remove the NOTE above?

schlosna:
@uschi2000 I could probably get numbers, but don't really have cycles to do so right now.

The JMH gradle plugin doesn't play well with nebula so adding microbenchmark in this project would be a pain, but I can try a separate project with this as a dependency if I get time.

hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
```diff
@@ -103,15 +103,25 @@ public boolean seekToNewSource(long targetPos) throws IOException {
...
+        if (bytesRead != -1) {
+            decryptedStreamPos += bytesRead;
+        }
+        return bytesRead;
     }
 
     @Override
     public void close() throws IOException {
         delegate.close();
+        decryptedStream.close();
```
uschi2000, line 124: should suffice to close the decryptedStream since it'll close the wrapped delegate

hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
schlosna: done

hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
schlosna: We could split this out to a separate PR, but it seems incorrect that we're not closing the CipherInputStream which finalizes the Cipher.

schlosna:
Did some quick non-JMH tests and it shows a ~30% speed up for files of size 128MB (typical Hadoop block size).

hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
uschi2000: ok, so let's close both here. i believe that `delegate.close()` is redundant.

uschi2000:
👍

uschi2000 (approved):

schlosna:
@uschi2000 I don't have write access to this to merge, so can you or another please merge at will and ideally tag a release to pickup downstream?

uschi2000:
Defer to Joe.

hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
ellisjoe: just close the decryptedStream. it'll take care of figuring out whether or not to close the delegate.

hadoop-crypto/src/test/java/com/palantir/hadoop/EncryptedFileSystemTest.java
```diff
@@ -104,7 +105,7 @@ public void testDelegateStreamIsClosed() throws IOException {
         myIs.close();
         fs.close();
 
-        verify(is).close();
+        verify(is, atLeastOnce()).close();
```
ellisjoe, line 108: remove this once you're closing only the decryptedStream

ellisjoe:
need to test the new method

schlosna:
EncryptedFileSystemTest implicitly tests this as you can tell from coverage report, but can add some unit tests verifying individual behavior and and finalization of cipher.

hadoop-crypto/src/main/java/com/palantir/hadoop/cipher/FsCipherInputStream.java
schlosna: Closing both is safe, but can adjust next week

PR #19, opened 2016-11-10
Add tests for bulk reads

schlosna:


ellisjoe:
don't think we can close the CipherInputStream in the close method of FsCipherInputStream because it will break any time we try to close it on a CipherStream that uses padding and doesn't isn't within 512 bytes of the end of the stream. The issue is that the CipherInputStream buffers up to 512 bytes from the underlying stream and that's what it passes to the underlying Cipher. When doFinal is called on the Cipher it only knows about what the CipherStream has passed it which doesn't include the end of the InputStream.

CipherStream buffer: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7u40-b43/javax/crypto/CipherInputStream.java#77
CipherStream filling buffer: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7u40-b43/javax/crypto/CipherInputStream.java#103
CipherStream updating Cipher: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7u40-b43/javax/crypto/CipherInputStream.java#120

and a test directly against cipher streams that passes when the byte array has length 511 but fails with length 512:

```
    @Test
    public final void test() throws NoSuchPaddingException, NoSuchAlgorithmException, IOException {
        ByteArrayOutputStream os = new ByteArrayOutputStream();

        Cipher encryptCipher = seekableCipher.initCipher(Cipher.ENCRYPT_MODE);
        CipherOutputStream outputStream = supplier.getOutputStream(os, encryptCipher);

        byte[] rand = new byte[511];
        random.nextBytes(rand);
        outputStream.write(rand);
        outputStream.close();

        Cipher decryptCipher = seekableCipher.initCipher(Cipher.DECRYPT_MODE);
        CipherInputStream inputStream = supplier.getInputStream(new ByteArrayInputStream(os.toByteArray()), decryptCipher);

        inputStream.read();
        inputStream.close();
    }
```

I don't think it worth trying to close the CipherInputStream in any case. If we relied on auth tags from something the GCM mode then maybe, but that validation only works when you read the entire encrypted stream sequentially anyway.

schlosna:
The `java.io.IOException: javax.crypto.BadPaddingException: Given final block not properly padded` on `CipherInputStream#close` is actually a JDK bug https://bugs.openjdk.java.net/browse/JDK-8064546 that was fixed in 7u85+ & 8u51+. Looks like we're using `oraclejdk7`, so we only get the last public Java 7 release 7u80 :(

I can adjust the tests to pass on broken JDK, but I still think we should be closing the `decryptedStream` since that's the one created in `FsCipherInputStream` so owns the resource lifecycle.

ellisjoe (approved):

PR #26, opened 2016-12-07
Move DecryptingSeekableInput to crypto-core

ellisjoe:


ellisjoe:
@markelliot @uschi2000 this gets you a `DecryptingInputStream` without any hadoop dependencies in the crypto-core project. Still need to move/rewrite the tests since they depend on `FSDataInputStream`.

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+package com.palantir.crypto.io;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.io.ByteStreams;
+import com.palantir.crypto.cipher.CipherStreamSupplier;
+import com.palantir.crypto.cipher.CipherStreamSupplierImpl;
+import com.palantir.crypto.cipher.SeekableCipher;
+import com.palantir.seekio.SeekableInput;
```
markelliot, line 24: it occurs to me that we're going to have to publish this jar to bintray, and it'll have to have a common group with the rest of this stuff =/

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -0,0 +1,116 @@
...
+        } else {
+            prevBlock = pos / blockSize - 1;
+            bytesToSkip = (int) (pos % blockSize + blockSize);
+        }
+
+        long prevBlockOffset = prevBlock * blockSize;
+        Cipher cipher = seekableCipher.seek(prevBlockOffset);
+        delegate.seek(prevBlockOffset);
+
+        // Need a new cipher stream since seeking the stream and cipher invalidate the cipher stream's buffer
```
markelliot, line 78: an optimization we could make (worth leaving a note?) is deciding if the seek is within a threshold, and if so, just reading the bytes rather than renewing the cipher

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -0,0 +1,116 @@
...
+        // causes "java.io.IOException: javax.crypto.BadPaddingException: Given final block not properly padded" and
+        // is not fixed until Java 7u85 (not publicly available) and Java 8u51.
+        // decryptedStream.close();
+    }
+
+    /**
+     * Convenience method for converting {@link SeekableInput} into a {@link InputStream}.
+     */
+    private static InputStream wrap(SeekableInput input) {
+        return new DefaultSeekableInputStream(input);
```
markelliot, line 113: it'd be cool if we could avoid the `InputStream` stuff

markelliot (approved):
left some ideas

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
markelliot: I guess we can't since the built in Cipher stuff takes an `InputStream`, shrug.

PR #31, opened 2016-12-15
Slightly improve object allocation

markelliot:


crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -56,6 +55,13 @@ public DecryptingSeekableInput(SeekableInput delegate, SeekableCipher cipher) {
      */
     @Override
     public void seek(long pos) throws IOException {
+        if (pos == getPos()) {
+            // short-circuit if no work to do
+            return;
+        }
+        // TODO the object allocation here has non-zero cost, for frequent skipping it may be cheaper to eat
```
uschi2000: nit: TODO(markelliot)

uschi2000 (commented):
Joe should check since this potentially alters behaviour.

ellisjoe (approved):

PR #32, opened 2016-12-16
Do less work during seeks by keeping some state

markelliot:


crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -32,8 +32,12 @@
     private final SeekableCipher seekableCipher;
     private final CipherStreamSupplier supplier;
 
-    private CipherInputStream decryptedStream;
-    private long decryptedStreamPos;
+    // block info
+    private long blockIndex = -1;
```
uschi2000, line 36: should add some comment here to clarify what exactly these are, invariants,e tc.

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -55,37 +57,29 @@ public DecryptingSeekableInput(SeekableInput delegate, SeekableCipher cipher) {
...
+        if (pos == decryptedStreamPos) {
             // short-circuit if no work to do
             return;
         }
-        // TODO(markelliot) the object allocation here has non-zero cost, for frequent skipping it may be cheaper to eat
-        // and throw away bytes between the current position and a positive next position
-
-        int blockSize = seekableCipher.getBlockSize();
-
-        // If pos is in the first block then seek to 0 and skip pos bytes
```
uschi2000, line 67: should keep this explanation somewhere

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -55,37 +57,29 @@ public DecryptingSeekableInput(SeekableInput delegate, SeekableCipher cipher) {
...
-        decryptedStream = supplier.getInputStream(delegate, cipher);
+            Cipher cipher = seekableCipher.seek(currentBlockStartRow);
+            delegate.seek(currentBlockStartRow);
+
+            currentBlockDecryptionStream = supplier.getInputStream(delegate, cipher);
+        }
 
         // Skip to the byte offset in the block where 'pos' is located
-        ByteStreams.skipFully(decryptedStream, bytesToSkip);
+        int toSkip = (int) (pos - currentBlockStartRow);
```
uschi2000, line 80: that cast looks potentially dangerous. are you confident you won't ever overflow?

markelliot: It's bounded by blockSize, which is integer sized, so it's safe. Will add a comment

schlosna: Could use Guava's `Ints.checkedCast(pos - currentBlockStartRow);` to assert that invariant

markelliot: this is knowably correct by inspection.

markelliot:
going to take another pass at this, looks like most of the tests were actually in hadoop-crypto and this makes a bunch of them fail.

two things I'm trying to optimize:
- negative seeks in the underlying SeekableInput (Hadoop or whatever) are pretty much always expensive, so we should try to avoid them
- seeking forward within a block shouldn't cause us to seek to the beginning of the block first (since while we assume seeks are cheap, not seeking is still cheaper)

PR #35, opened 2016-12-21
[WIP] sr/native libs

sjrand:
Same idea as #22, which I didn't actually mean to close.


sjrand:
Finally mostly ready for review. There are still 2 failing tests that I need to fix in `FsCipherOutputStreamTest`, but I'm pretty sure those are a consequence of me being bad at mocking, rather than correctness problems. Will try to get to those over the next couple of days.

cc @ellisjoe @schlosna @pwoody for when break is over

Some stats from brief, not particularly rigorous testing on a 3-node YARN cluster (need to find a better way of doing this):

* Reading 10g of Parquet from `ehdfs` via Spark [1]:

   * 2.0.0-rc3: 84 seconds
   * this branch without libcrypto.so on NodeManagers' LD_LIBRARY_PATH: 84 seconds
   * this branch with libcrypto.so on NodeManagers' LD_LIBRARY_PATH: 56 seconds

* Writing 10g of parquet into `ehdfs` via Spark [2]:

   * 2.0.0-rc3: 209 seconds
   *  this branch without libcrypto.so on NodeManagers' LD_LIBRARY_PATH: 120 seconds (I'm confused by this, expected same as 2.0.0-rc3)
   * this branch with libcrypto.so on NodeManagers' LD_LIBRARY_PATH: 96 seconds

[1] `spark.read.parquet("ehdfs://<path_to_data>").cache().count()`
[2] `sc.parallelize(1 to (10*1024*1024)).map(k => Seq(k, org.apache.commons.lang.RandomStringUtils.random(1024, "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWxyZ0123456789")).mkString("|")).toDF().write.parquet("ehdfs://<path_to_data>")`

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -115,7 +146,7 @@ public long getPos() throws IOException {
     @Override
     public int read(byte[] buffer, int offset, int length) throws IOException {
         int bytesRead = decryptedStream.read(buffer, offset, length);
-        if (bytesRead != -1) {
+        if (bytesRead != 0) {
```
schlosna: This seems off, -1 indicates end of stream so The previous should be proper behavior.

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/AbstractFsCipherInputStreamTest.java
```diff
@@ -165,8 +166,11 @@ public final void testBulkRead() throws IOException {
         byte[] buffer = new byte[NUM_BYTES];
         int offset = 0;
         int read;
-        while ((read = cis.read(buffer, offset, buffer.length)) != -1) {
+        int len = buffer.length;
+        while ((read = cis.read(buffer, offset, len)) != 0) {
             offset += read;
+            len -= read;
```
schlosna: Why do we adjust the len here? If we didn't read the full buffer length, the next time through should try to read from offset to buffer.length

schlosna:
I'll take a more in depth look in a week.

crypto-core/src/main/java/com/palantir/crypto/cipher/AesCtrCipher.java
```diff
@@ -89,21 +86,20 @@ public Cipher seek(long pos) {
...
             cipher.init(currentOpmode, key, newIv);
-        } catch (InvalidKeyException | InvalidAlgorithmParameterException e) {
+
+            // Skip to the byte offset in the block where 'pos' is located
+            int bytesToSkip = (int) (pos % BLOCK_SIZE);
+            byte[] skip = new byte[bytesToSkip];
+            byte[] out = new byte[BLOCK_SIZE];
+            cipher.update(skip, 0, bytesToSkip, out, 0);
+            return cipher;
+        } catch (GeneralSecurityException e) {
```
uschi2000: can catch more narrowly?

crypto-core/src/main/java/com/palantir/crypto/cipher/AesCtrCipher.java
```diff
@@ -89,21 +86,20 @@ public Cipher seek(long pos) {
...
-        // Init the cipher with the new iv
         try {
+            CryptoCipher cipher = CryptoCipherFactory.getCryptoCipher(ALGORITHM);
             cipher.init(currentOpmode, key, newIv);
-        } catch (InvalidKeyException | InvalidAlgorithmParameterException e) {
+
+            // Skip to the byte offset in the block where 'pos' is located
+            int bytesToSkip = (int) (pos % BLOCK_SIZE);
+            byte[] skip = new byte[bytesToSkip];
+            byte[] out = new byte[BLOCK_SIZE];
```
uschi2000: can use static memory for this, right?

crypto-core/src/main/java/com/palantir/crypto/cipher/AesCtrCipher.java
```diff
@@ -89,21 +86,20 @@ public Cipher seek(long pos) {
...
+            cipher.update(skip, 0, bytesToSkip, out, 0);
```
uschi2000: would be nice do describe the zeros inline: `update(skip, 0 /* what? */, bytesToSkip, out, 0 /* what? */)`

crypto-core/src/main/java/com/palantir/crypto/cipher/CipherStreamSupplier.java
```diff
@@ -30,12 +35,14 @@
     /**
      * Wraps the {@link InputStream} in a {@link CipherInputStream} which uses the given {@link Cipher} for decryption.
```
uschi2000: stale docs; need to describe what all those parameters are.

crypto-core/src/main/java/com/palantir/crypto/cipher/CipherStreamSupplier.java
```diff
@@ -30,12 +35,14 @@
...
-    CipherInputStream getInputStream(InputStream is, Cipher cipher);
+    CryptoInputStream getInputStream(String transformation, Properties props, InputStream is,
+            Key key, AlgorithmParameterSpec params);
 
     /**
      * Wraps the {@link OutputStream} in a {@link CipherOutputStream} which uses the given {@link Cipher} for
      * encryption.
      */
-    CipherOutputStream getOutputStream(OutputStream os, Cipher cipher);
+    CryptoOutputStream getOutputStream(String transformation, Properties props, OutputStream os,
```
uschi2000: as above

crypto-core/src/main/java/com/palantir/crypto/cipher/CipherStreamSupplier.java
```diff
@@ -30,12 +35,14 @@
     /**
      * Wraps the {@link InputStream} in a {@link CipherInputStream} which uses the given {@link Cipher} for decryption.
      */
-    CipherInputStream getInputStream(InputStream is, Cipher cipher);
+    CryptoInputStream getInputStream(String transformation, Properties props, InputStream is,
```
uschi2000: why do we need all these parameters?

crypto-core/src/main/java/com/palantir/crypto/cipher/SeekableCipher.java
```diff
@@ -18,24 +18,26 @@
 
 import com.palantir.crypto.KeyMaterial;
 import javax.crypto.Cipher;
+import org.apache.commons.crypto.cipher.CryptoCipher;
 
 /**
- * Provides access to a {@link Cipher} with the ability to {@link #seek} the Cipher.
+ * Provides access to a {@link org.apache.commons.crypto.cipher.CryptoCipher}
```
uschi2000: don't need to fully qualify

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -23,34 +23,47 @@
...
         this.supplier = supplier;
+        this.secretKey = cipher.getKeyMaterial().getSecretKey();
+        this.ivParameterSpec = new IvParameterSpec(cipher.getKeyMaterial().getIv());
 
         // this bound could be tightened just 2x block size, but we pad a bit because reading and decrypting a few
         // extra bytes is likely to be less costly than object allocations and work associated with a forward seek
         this.skipThreshold = seekableCipher.getBlockSize() * 4;
 
-        decryptedStream = supplier.getInputStream(delegate, cipher.initCipher(Cipher.DECRYPT_MODE));
+        decryptedStream = supplier.getInputStream(transformation, new Properties(),
```
uschi2000: looks like the Properties can be removed from the signature?

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -23,34 +23,47 @@
...
+    private final IvParameterSpec ivParameterSpec;
+    static final int IV_SIZE = 16;
 
-    private CipherInputStream decryptedStream;
+    private CryptoInputStream decryptedStream;
     private long decryptedStreamPos;
 
-    public DecryptingSeekableInput(SeekableInput delegate, SeekableCipher cipher) {
-        this(delegate, cipher, new CipherStreamSupplierImpl());
+    public DecryptingSeekableInput(String transformation, SeekableInput delegate, SeekableCipher cipher) {
```
uschi2000: needs docs

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -23,34 +23,47 @@
...
 
 public final class DecryptingSeekableInput implements SeekableInput {
 
     private final DefaultSeekableInputStream delegate;
     private final SeekableCipher seekableCipher;
     private final CipherStreamSupplier supplier;
     private final long skipThreshold;
+    private final Key secretKey;
+    private final IvParameterSpec ivParameterSpec;
+    static final int IV_SIZE = 16;
```
uschi2000: IV_SIZE defined in multiple places, something's off here with the decomp

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
```diff
@@ -96,11 +109,29 @@ public void seek(long pos) throws IOException {
         }
 
         long prevBlockOffset = prevBlock * blockSize;
-        Cipher cipher = seekableCipher.seek(prevBlockOffset);
+        CryptoCipher cipher = seekableCipher.seek(prevBlockOffset);
         delegate.seek(prevBlockOffset);
 
+        // TODO(srand) This is duplicating work already done in AesCtrCipher#seek as a result of the public
```
uschi2000: I believe we should do that before merging this PR. maybe submit a pure refactor PR that sets up the signatures you need.

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
uschi2000: is there a test for this? (there should be)

crypto-core/src/test/java/com/palantir/crypto/cipher/AbstractSeekableCipherTest.java
```diff
@@ -126,12 +132,32 @@ public final void testGetKeyMaterial() {
...
+    public final void testEncryptDecrypt(CryptoCipher encryptingCipher, CryptoCipher decryptingCipher)
             throws BadPaddingException, IllegalBlockSizeException {
         byte[] data = new byte[NUM_BLOCKS * encryptingCipher.getBlockSize()];
         random.nextBytes(data);
-        byte[] encryptedData = encryptingCipher.doFinal(data);
-        byte[] decryptedData = decryptingCipher.update(encryptedData);
+
+        // Account for padding or the lack thereof.
+        byte[] encryptedData;
+        if (getAlgorithm().equals(AesCtrCipher.ALGORITHM)) {
```
uschi2000: decomp is a little weird here. don't know a better way thogh.

crypto-core/src/test/java/com/palantir/crypto/cipher/AbstractSeekableCipherTest.java
```diff
@@ -126,12 +132,32 @@ public final void testGetKeyMaterial() {
...
+            encryptedData = new byte[encryptingCipher.getBlockSize() * NUM_BLOCKS];
+        } else if (getAlgorithm().equals(AesCbcCipher.ALGORITHM)) {
+            encryptedData = new byte[encryptingCipher.getBlockSize() * (NUM_BLOCKS + 1)];
+        } else {
+            throw new IllegalArgumentException("Must specify either \"AES/CBC/PKCS5Padding\""
+                    + "or \"AES/CTR/NoPadding\"");
+        }
+        try {
+            encryptingCipher.doFinal(data, 0, data.length, encryptedData, 0);
+        } catch (ShortBufferException e) {
```
uschi2000: just make the test throw

crypto-core/src/test/java/com/palantir/crypto/cipher/AesCtrCipherTest.java
```diff
@@ -82,9 +88,14 @@ public void testNistExample(int opmode, int blockNumber, String input, String ou
         KeyMaterial keyMaterial = KeyMaterial.of(new SecretKeySpec(key, AesCtrCipher.KEY_ALGORITHM), iv);
         SeekableCipher seekableCipher = getCipher(keyMaterial);
         seekableCipher.initCipher(opmode);
-        Cipher cipher = seekableCipher.seek(blockNumber * (long) AesCtrCipher.BLOCK_SIZE);
+        CryptoCipher cipher = seekableCipher.seek(blockNumber * (long) AesCtrCipher.BLOCK_SIZE);
 
-        byte[] finalBytes = cipher.update(inputBytes);
+        byte[] finalBytes = new byte[outputBytes.length];
+        try {
```
uschi2000: make test throw

crypto-core/src/test/java/com/palantir/crypto/cipher/CipherStreamSupplierImplTest.java
```diff
@@ -33,24 +39,40 @@
...
         supplier = new CipherStreamSupplierImpl();
+        byte[] iv = new byte[16];
+        SecureRandom rng = new SecureRandom();
+        rng.nextBytes(iv);
+        this.ivParameterSpec = new IvParameterSpec(iv);
+        try {
+            KeyGenerator keyGen = KeyGenerator.getInstance("AES");
+            keyGen.init(256);
+            this.key = keyGen.generateKey();
+        } catch (NoSuchAlgorithmException e) {
```
uschi2000: make method throw

crypto-serialization/src/main/java/com/palantir/crypto/KeyMaterials.java
```diff
@@ -42,6 +44,15 @@
 
     private static final Logger log = LoggerFactory.getLogger(KeyMaterials.class);
     private static final Map<Integer, ? extends KeySerializer> SERIALIZERS = KeySerializers.getSerializers();
+    private static final CryptoRandom random = getRandom();
+
+    private static CryptoRandom getRandom() {
+        try {
+            return CryptoRandomFactory.getCryptoRandom();
```
uschi2000: is that using the correct entropy source?

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/AbstractFsCipherInputStreamTest.java
```diff
@@ -165,8 +166,11 @@ public final void testBulkRead() throws IOException {
         byte[] buffer = new byte[NUM_BYTES];
         int offset = 0;
         int read;
-        while ((read = cis.read(buffer, offset, buffer.length)) != -1) {
+        int len = buffer.length;
+        while ((read = cis.read(buffer, offset, len)) != 0) {
```
uschi2000: 0 is a perfectly continuation case for an InputStream; should terminate when reading -1

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/AbstractFsCipherInputStreamTest.java
uschi2000: +1. just always read at most BUFSIZE, no need to adjust this.

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/AesCbcCipherInputStreamTest.java
```diff
@@ -23,9 +23,14 @@
 public final class AesCbcCipherInputStreamTest extends AbstractFsCipherInputStreamTest {
 
     @Override
-    SeekableCipher getSeekableCipher() {
+    SeekableCipher  getSeekableCipher() {
```
uschi2000: oops

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/FsCipherOutputStreamTest.java
```diff
@@ -21,49 +21,79 @@
...
-        cos = mock(CipherOutputStream.class);
+        cos = mock(CryptoOutputStream.class);
         seekableCipher = mock(SeekableCipher.class);
         supplier = mock(CipherStreamSupplier.class);
+        keyMaterial = mock(KeyMaterial.class);
+        ivParameterSpec = mock(IvParameterSpec.class);
+        byte[] iv = new byte[16];
+        SecureRandom rng = new SecureRandom();
+        rng.nextBytes(iv);
+        // this.ivParameterSpec = new IvParameterSpec(iv);
```
uschi2000: oops

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/FsCipherOutputStreamTest.java
```diff
@@ -21,49 +21,79 @@
...
+        try {
+            KeyGenerator keyGen = KeyGenerator.getInstance("AES");
+            keyGen.init(256);
+            this.key = keyGen.generateKey();
+        } catch (NoSuchAlgorithmException e) {
```
uschi2000: make method throw

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/FsCipherOutputStreamTest.java
```diff
@@ -21,49 +21,79 @@
...
+            throw Throwables.propagate(e);
+        }
 
         when(seekableCipher.initCipher(anyInt())).thenReturn(initCipher);
-        when(supplier.getOutputStream(os, initCipher)).thenReturn(cos);
+        when(supplier.getOutputStream(transformation, new Properties(), os,
+                key, ivParameterSpec)).thenReturn(cos);
+        when(seekableCipher.getKeyMaterial()).thenReturn(keyMaterial);
+        when(keyMaterial.getIv()).thenReturn(iv);
```
uschi2000: probably easier to use a real keymaterial

uschi2000 (changes requested):
- This is a pretty big change; is it possible to split this up into smaller chunks?
- Should have compatibility tests: decrypt_old(encrypt_new(data)) == data, and decrypt_new(encrypt_old(data)) == data
- Assuming the purpose of this change is to allow native crypt libraries to be used, the README needs to be updated with instructions on how to do that.
- Further, should we also have a log statement somewhere that gives feedback on what crypto libraries exactly are used?

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
sjrand: Yep, good call, I wasn't reading `CryptoInputStream#read` carefully enough. Will revert and add a test

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/AbstractFsCipherInputStreamTest.java
sjrand: I made the change because the test was failing with an `IndexOutOfBoundsException` from https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java#L244. We were hitting the `len > b.length - off` case that it checks for. Is there a better way to handle that?

crypto-core/src/main/java/com/palantir/crypto/cipher/AesCtrCipher.java
sjrand: Reason for the `GeneralSecurityException` is that it's what `CryptoCipherFactory.getCryptoCipher()` throws, and then the exceptions that the other methods in the try block throw are all subclasses of that.

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
sjrand: good call, will do

crypto-core/src/main/java/com/palantir/crypto/cipher/CipherStreamSupplier.java
sjrand: I added those to be able to construct a `CryptoInputStream`, since those are things that the public constructor for that class wants. It does make this change significantly more invasive and messy though, so definitely open to changing my approach if there's a better way.

crypto-core/src/main/java/com/palantir/crypto/cipher/CipherStreamSupplier.java
sjrand: same as above

crypto-core/src/main/java/com/palantir/crypto/cipher/CipherStreamSupplier.java
sjrand: (though I did get rid of the `Properties` param per your comment below)

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/AbstractFsCipherInputStreamTest.java
sjrand: Something is weird here. You guys are right, this should be `-1`, but that makes the test get stuck in an infinite loop and my laptop hum increasingly loudly.

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
sjrand: Would this be adding a method to `SeekableCipher` that returns the IV of the block at which we're going to start decrypting? Then each implementation could do something like:

```
private byte[] currIv;
public byte[] getCurrentIv() { return currIv; }

// In the constructor:
currIv = initIv;

// And in the seek method, use the code that already exists to update the value of currIv
```

Does that make sense, or did you have something entirely different in mind?

crypto-serialization/src/main/java/com/palantir/crypto/KeyMaterials.java
sjrand: If the native crypto libs are available, and the hardware has the `RDRAND` instruction, it should use that. Otherwise, I gather from https://wiki.openssl.org/index.php/Random_Numbers that it will use `/dev/urandom` if the native libs are present. Without native libs, it'll just use a Java `SecureRandom`, presumably using whatever the default entropy source of the system is.

We could probably call `getCryptoRandom(properties)` where `properties` specifies the entropy source for Java to use. My understanding is that `/dev/urandom` is correct, and `/dev/random` is not (since it blocks)?

crypto-core/src/main/java/com/palantir/crypto/cipher/AesCtrCipher.java
uschi2000: oh, things like `update()` throw too? cool.

crypto-core/src/main/java/com/palantir/crypto/cipher/CipherStreamSupplier.java
uschi2000: what does `transformation` mean, for instance? would prefer to define this interface as narrowly as possible and hide away the somewhat ugly API for the underlying apache commons class.

crypto-core/src/main/java/com/palantir/crypto/io/DecryptingSeekableInput.java
uschi2000: @ellisjoe probably has the best intuition here.

crypto-serialization/src/main/java/com/palantir/crypto/KeyMaterials.java
uschi2000: Yep, `/dev/urandom` is what we want and Java SecureRandom is a good fallback.

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/AbstractFsCipherInputStreamTest.java
uschi2000: Bug in the apache libs?

sjrand:
Answers to questions:

* This is a pretty big change; is it possible to split this up into smaller chunks?

Could easily split the SecureRandom -> CryptoRandom change in `KeyMaterials` out, but that's so small, not sure it's worth it. Aside from that, I can't think of how I would break it up into parts. The two main components of the change are `Cipher` -> `CryptoCipher` and `Cipher{Input|Output}Stream` -> `Crypto{Input|Output}Stream`, and I don't think it would make sense to do one without the other. Definitely open to suggestions though.

* Should have compatibility tests: decrypt_old(encrypt_new(data)) == data, and decrypt_new(encrypt_old(data)) == data

Yep, makes sense

* Assuming the purpose of this change is to allow native crypt libraries to be used, the README needs to be updated with instructions on how to do that.

Indeed it should

* Further, should we also have a log statement somewhere that gives feedback on what crypto libraries exactly are used?

Yep, that would be quite useful

So things to do (so that I don't forget):

* Add test for DecryptingSeekableInput#read
* Add docs to DecryptingSeekableInputStream constructor
* Figure out what's wrong with AbstractFsCipherInputStreamTest#testBulkRead. You guys are right that it should be `while ((read = cis.read(buffer, offset, len)) != -1)`, but for some reason the `read` method is never returning -1.
* Refactor (possibly in separate PR?) such that we don't compute the new IV twice after seeking.
* Try to clean up the CipherStreamSupplier interface to not require so many things
* Add compatibility tests as described above
* Update README with how to deploy native libs
* Log which libraries/classes are being used for crypto

crypto-core/src/main/java/com/palantir/crypto/cipher/CipherStreamSupplier.java
sjrand: The `transformation` is the algorithm, e.g., `"AES/CBC/PKCS5Padding"`

One idea is to make it take an `InputStream` and a `SeekableCipher`, and then get the `Key` and `AlgorithmParameterSpec` that we need from the SeekableCipher. Could also get the algorithm from the `SeekableCipher` if we were to add a `getAlgorithm()` method to it, which seems reasonable enough to me. Would that be better?

hadoop-crypto/src/test/java/com/palantir/hadoop/cipher/AbstractFsCipherInputStreamTest.java
sjrand: Gah. It's b/c we're decrementing `len`, so when `len == 0` we hit https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java#L246

uschi2000:
Cool, so not decrementing len should fix this, right?

sjrand:
Not decrementing `len` has the problem where you hit https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java#L244 and get an `IndexOutOfBoundsException`.

The last commit takes a similar approach to https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FSInputStream.java#L116 which seems reasonable enough.

crypto-core/src/main/java/com/palantir/crypto/cipher/AesCtrCipher.java
```diff
@@ -37,11 +35,11 @@
 public final class AesCtrCipher implements SeekableCipher {
 
     public static final String ALGORITHM = "AES/CTR/NoPadding";
-    static final String PROVIDER = "SunJCE";
     static final String KEY_ALGORITHM = "AES";
     static final int KEY_SIZE = 256;
     static final int BLOCK_SIZE = 16;
     static final int IV_SIZE = 16;
+    static byte[] out = new byte[BLOCK_SIZE];
```
markelliot: this looks like a potential place to run into some concurrency issues, I think either make this non-static (and either note that this class isn't thread-safe or mark the `seek` method synchronized) or make this `ThreadLocal`.

markelliot: actually, this never appears to be read, why do we need it at all?

ellisjoe:
Before digging into this too much, what is the expected benefit of switching from Java to Apache crypto libs? Are there any benchmarks to look at?

sjrand:
@ellisjoe, yep, some informal testing has been promising. See https://github.com/palantir/hadoop-crypto/pull/35#issuecomment-268874241 for what I got with the initial version of this PR. Not related to this library, but similarly, I saw a write to HDFS from a single Spark executor (2 gigs of data) go from 80 seconds to 59 seconds after adding `libcrypto.so` to the NodeManagers' library paths.

schlosna:
I haven't gotten back to analyze any of this yet, but wanted to add a couple related links I ran across recently: https://calomel.org/aesni_ssl_performance.html
https://twitter.com/a_z_e_t/status/819696277827022849

crypto-core/src/main/java/com/palantir/crypto/cipher/AesCtrCipher.java
sjrand: As far as I can tell it's needed for the call to `cipher.update(skip, 0 /* inputOffset */, bytesToSkip, out, 0 /* outputOffset */);` below. As you say we don't use `out`, so it'd be good if we could find a way to not initialize it, but my reading of https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/cipher/CryptoCipher.java is that we need to write to something for `update` to work.

sjrand:
After a brief four-month delay, I'm actively working on this again. Finally have all tests passing for the first time. This latest attempt is a bit less invasive w.r.t method signatures -- instead of dramatically changing the inputs to the methods in `CipherStreamSupplier`, we just have it take a `SeekableCipher` instead of the underlying `CryptoCipher`, with maybe a bit of weirdness in that the underlying cipher has to be initialized first. This results in a much smaller diff to the production code, though some invasive changes to the tests still seem necessary.

@uschi2000, I still owe you compatibility tests and logging for which underlying libs will be used (JCE vs openssl).

And of course there's the obvious question of what difference does this make to perf -- open to ideas on how I can test this somewhat rigorously. In the past I've just opened up a spark-shell, but it'd be great if we could have something more automated. Happy to try to contribute that if someone can point me in the right direction.

crypto-core/src/main/java/com/palantir/crypto2/io/DecryptingSeekableInput.java
```diff
@@ -24,19 +24,19 @@
...
 import javax.crypto.Cipher;
-import javax.crypto.CipherInputStream;
+import org.apache.commons.crypto.stream.CryptoInputStream;
 
 public final class DecryptingSeekableInput implements SeekableInput {
 
-    /** Size of the {@link CipherInputStream} internal buffer. */
-    private static final int CIPHER_INPUT_STREAM_BUFFER_SIZE = 512;
+    /** Size of the {@link CryptoInputStream} internal buffer. */
+    private static final int CRYPTO_INPUT_STREAM_BUFFER_SIZE = 8192;
```
sjrand, line 31: Unlike `CipherInputStream`, `CryptoInputStream` actually allows us to configure this. The default happens to be 8192, but it's likely worth playing with. I have a suspicion that we want it to align with `io.file.buffer.size`, but would need to read more code to be sure. @schlosna, interested to know if you have thoughts on how to determine the best value for this.

ellisjoe:
Ran some tests locally and this definitely looks promising. I'm gonna take a look at how we can get this in minimally without breaking all of our apis.

Results:
```
Benchmark                          Mode  Cnt  Score   Error  Units
BenchmarkReadWrite.apacheFileRead    ss   10  0.216 ± 0.041   s/op
BenchmarkReadWrite.apacheMemRead     ss   10  0.167 ± 0.038   s/op
BenchmarkReadWrite.javaFileRead      ss   10  0.569 ± 0.030   s/op
BenchmarkReadWrite.javaMemRead       ss   10  0.442 ± 0.038   s/op
BenchmarkReadWrite.fileRead          ss   10  0.172 ± 0.050   s/op
BenchmarkReadWrite.memRead           ss   10  0.117 ± 0.007   s/op
```

sjrand:
@ellisjoe , thanks for taking a look. I'm curious, in your local tests, did you have the native libs on `LD_LIBRARY_PATH` or `java.library.path`? (Assuming yes, but wanted to confirm.)

ellisjoe:
FYI https://github.com/palantir/hadoop-crypto/pull/61

had to set `LD_LIBRARY_PATH` and set it to `/usr/local/opt/openssl/lib` which is the brew installed version of ssl, not the mac version.

sjrand:
word, thanks

sjrand:
(Also happy to do the refactor if there are ideas for how to do this less disruptively.)

sjrand:
@uschi2000 I've now added the compatibility test and the logging

crypto-core/src/main/java/com/palantir/crypto2/cipher/SeekableCipher.java
```diff
@@ -18,33 +18,47 @@
...
  */
 public interface SeekableCipher {
 
     /**
-     * Initialize the underlying {@link Cipher} to either {@link Cipher#ENCRYPT_MODE} or {@link Cipher#DECRYPT_MODE}.
+     * Initialize the underlying {@link CryptoCipher} to either {@link Cipher#ENCRYPT_MODE} or
+     * {@link Cipher#DECRYPT_MODE}.
      */
-    Cipher initCipher(int opmode);
+    CryptoCipher initCipher(int opmode);
```
uschi2000: note that this change requires a new release (and likely maven&java name).

crypto-core/src/main/java/com/palantir/crypto2/cipher/AesCtrCipher.java
```diff
@@ -37,37 +35,39 @@
 public final class AesCtrCipher implements SeekableCipher {
 
     public static final String ALGORITHM = "AES/CTR/NoPadding";
-    static final String PROVIDER = "SunJCE";
     static final String KEY_ALGORITHM = "AES";
     static final int KEY_SIZE = 256;
     static final int BLOCK_SIZE = 16;
     static final int IV_SIZE = 16;
+    private static byte[] out = new byte[BLOCK_SIZE];
```
uschi2000: we're never reading this, right? maybe indicate this in the name and/or comment.

crypto-core/src/main/java/com/palantir/crypto2/cipher/AesCtrCipher.java
```diff
@@ -80,30 +80,27 @@ public Cipher seek(long pos) {
...
         // Init the cipher with the new iv
         try {
-            cipher.init(currentOpmode, key, newIv);
-        } catch (InvalidKeyException | InvalidAlgorithmParameterException e) {
+            CryptoCipher cipher = CryptoCipherFactory.getCryptoCipher(ALGORITHM);
+            cipher.init(currentOpmode, key, currIvParameterSpec);
+
+            // Skip to the byte offset in the block where 'pos' is located
+            int bytesToSkip = (int) (pos % BLOCK_SIZE);
+            byte[] skip = new byte[bytesToSkip];
```
uschi2000: should this also be a static variable, similar to `out`?

uschi2000: should also leave a comment explaining the relationship between the sizes of skip/out and BLOCK_SIZE.

crypto-core/src/main/java/com/palantir/crypto2/cipher/CipherStreamSupplierImpl.java
```diff
@@ -16,22 +16,40 @@
...
-        return new CipherInputStream(is, cipher);
+    public CryptoInputStream getInputStream(InputStream is, SeekableCipher seekableCipher) {
+        try {
+            return new CryptoInputStream(seekableCipher.getAlgorithm(),
+                    new Properties(),
+                    is,
+                    seekableCipher.getKeyMaterial().getSecretKey(),
+                    seekableCipher.getCurrIv());
+        } catch (IOException e) {
+            throw Throwables.propagate(e);
```
uschi2000: this is deprecated.

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/KeyMaterials.java
```diff
@@ -43,6 +45,15 @@
 
     private static final Logger log = LoggerFactory.getLogger(KeyMaterials.class);
     private static final Map<Integer, ? extends KeySerializer> SERIALIZERS = KeySerializers.getSerializers();
+    private static final CryptoRandom random = getRandom();
+
+    private static CryptoRandom getRandom() {
+        try {
+            return CryptoRandomFactory.getCryptoRandom();
+        } catch (GeneralSecurityException e) {
+            throw Throwables.propagate(e);
```
uschi2000: deprecated

crypto-core/src/main/java/com/palantir/crypto2/cipher/AesCbcCipher.java
```diff
@@ -45,22 +42,24 @@
...
+    public CryptoCipher initCipher(int opmode) {
         this.currentOpmode = opmode;
         try {
-            Cipher cipher = getInstance();
+            CryptoCipher cipher = CryptoCipherFactory.getCryptoCipher(ALGORITHM);
             cipher.init(opmode, key, new IvParameterSpec(initIv));
             return cipher;
-        } catch (InvalidKeyException | InvalidAlgorithmParameterException e) {
+        } catch (GeneralSecurityException e) {
             throw Throwables.propagate(e);
```
uschi2000: deprecated

crypto-core/src/main/java/com/palantir/crypto2/cipher/AesCtrCipher.java
```diff
@@ -37,37 +35,39 @@
...
+    public CryptoCipher initCipher(int opmode) {
         this.currentOpmode = opmode;
         try {
-            Cipher cipher = getInstance();
+            CryptoCipher cipher = CryptoCipherFactory.getCryptoCipher(ALGORITHM);
             cipher.init(opmode, key, new IvParameterSpec(initIv));
             return cipher;
-        } catch (InvalidKeyException | InvalidAlgorithmParameterException e) {
+        } catch (GeneralSecurityException e) {
             throw Throwables.propagate(e);
```
uschi2000: deprecated

crypto-core/src/main/java/com/palantir/crypto2/cipher/AesCtrCipher.java
```diff
@@ -80,30 +80,27 @@ public Cipher seek(long pos) {
...
+            CryptoCipher cipher = CryptoCipherFactory.getCryptoCipher(ALGORITHM);
+            cipher.init(currentOpmode, key, currIvParameterSpec);
+
+            // Skip to the byte offset in the block where 'pos' is located
+            int bytesToSkip = (int) (pos % BLOCK_SIZE);
+            byte[] skip = new byte[bytesToSkip];
+            cipher.update(skip, 0 /* inputOffset */, bytesToSkip, out, 0 /* outputOffset */);
+            return cipher;
+        } catch (GeneralSecurityException e) {
             throw Throwables.propagate(e);
```
uschi2000: deprecated

crypto-core/src/main/java/com/palantir/crypto2/cipher/CipherStreamSupplierImpl.java
```diff
@@ -16,22 +16,40 @@
...
-        return new CipherOutputStream(os, cipher);
+    public CryptoOutputStream getOutputStream(OutputStream os, SeekableCipher seekableCipher) {
+        try {
+            return new CryptoOutputStream(seekableCipher.getAlgorithm(),
+                    new Properties(),
+                    os,
+                    seekableCipher.getKeyMaterial().getSecretKey(),
+                    seekableCipher.getCurrIv());
+        } catch (IOException e) {
+            throw Throwables.propagate(e);
```
uschi2000: deprecated

crypto-core/src/main/java/com/palantir/crypto2/cipher/CipherStreamSupplierImpl.java
```diff
@@ -16,22 +16,40 @@
...
+        }
     }
 
     @Override
-    public CipherOutputStream getOutputStream(OutputStream os, Cipher cipher) {
-        return new CipherOutputStream(os, cipher);
+    public CryptoOutputStream getOutputStream(OutputStream os, SeekableCipher seekableCipher) {
+        try {
+            return new CryptoOutputStream(seekableCipher.getAlgorithm(),
+                    new Properties(),
```
uschi2000, line 45: are these safe to reuse?

crypto-core/src/main/java/com/palantir/crypto2/cipher/CipherStreamSupplierImpl.java
```diff
@@ -16,22 +16,40 @@
...
 
 public final class CipherStreamSupplierImpl implements CipherStreamSupplier {
 
     @Override
-    public  CipherInputStream getInputStream(InputStream is, Cipher cipher) {
-        return new CipherInputStream(is, cipher);
+    public CryptoInputStream getInputStream(InputStream is, SeekableCipher seekableCipher) {
+        try {
+            return new CryptoInputStream(seekableCipher.getAlgorithm(),
+                    new Properties(),
```
uschi2000, line 32: is there a constructor that doesn't use `Properties`?

crypto-core/src/main/java/com/palantir/crypto2/cipher/SeekableCipher.java
```diff
@@ -18,33 +18,47 @@
...
+    /**
+     * Returns the algorithm used by the underlying {@link CryptoCipher} for encryption and decryption.
+     */
+    String getAlgorithm();
+
+    /**
+     * Returns the initialization vector for the block that the underlying {@link CryptoCipher} will encrypt or decrypt
+     * next.
+     */
+    IvParameterSpec getCurrIv();
```
uschi2000, line 58: I don't quite understand why the apache API needs this additional information while the javax library does not. can you enlighten me?

crypto-core/src/test/java/com/palantir/crypto2/cipher/CipherStreamSupplierImplTest.java
```diff
@@ -33,24 +39,39 @@
...
-        cipher = mock(Cipher.class);
+        seekableCipher = mock(SeekableCipher.class);
         supplier = new CipherStreamSupplierImpl();
+
+        byte[] iv = new byte[16];
+        SecureRandom rng = new SecureRandom();
+        rng.nextBytes(iv);
+        KeyGenerator keyGen = KeyGenerator.getInstance("AES");
+        keyGen.init(256);
+        this.key = keyGen.generateKey();
```
uschi2000: no need for `this`

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/KeyMaterials.java
```diff
@@ -43,6 +45,15 @@
 
     private static final Logger log = LoggerFactory.getLogger(KeyMaterials.class);
     private static final Map<Integer, ? extends KeySerializer> SERIALIZERS = KeySerializers.getSerializers();
+    private static final CryptoRandom random = getRandom();
+
+    private static CryptoRandom getRandom() {
+        try {
+            return CryptoRandomFactory.getCryptoRandom();
```
uschi2000, line 52: I assume this initialized with a time/entropy-dependent seed?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -86,6 +90,13 @@ public void initialize(URI uri, Configuration conf) throws IOException {
         KeyPair keyPair = getKeyPair(conf);
         KeyStorageStrategy keyStore = new FileKeyStorageStrategy(delegate, keyPair);
 
+        if (Crypto.isNativeCodeLoaded()) {
```
uschi2000, line 93: do we have CI-automation in this project that runs tests etc with-vs-without native libraries?

uschi2000 (changes requested):

crypto-core/src/main/java/com/palantir/crypto2/cipher/CipherStreamSupplierImpl.java
sjrand: from what I can tell all the public constructors require a `Properties` argument

crypto-core/src/main/java/com/palantir/crypto2/cipher/CipherStreamSupplierImpl.java
sjrand: Is the idea that we could just have a single (static?) `Properties` object, and use that each time?

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/KeyMaterials.java
sjrand: It'll default to using an `OpenSslCryptoRandom`, which'll use hardware instructions if possible, and if not, then it looks like it'll use `/dev/random` based on my reading of https://wiki.openssl.org/index.php/Random_Numbers. If native libs aren't loaded, it'll fall back to using a `JavaCryptoRandom`, which uses a `java.security.SecureRandom` under the hood.

sjrand: I _think_ that in both cases the answer to your question is yes.

crypto-core/src/main/java/com/palantir/crypto2/cipher/SeekableCipher.java
sjrand: Is that because it's changing the return value of a method in a public interface?

crypto-core/src/main/java/com/palantir/crypto2/cipher/AesCtrCipher.java
sjrand: Is it safe to make `skip` static given that its length will change depending on the value of `bytesToSkip`?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
sjrand: with the idea being that we could write a unit to make that this logging works as intended?

Definitely seems useful, both for this and in general. I'll investigate.

crypto-core/src/main/java/com/palantir/crypto2/cipher/SeekableCipher.java
sjrand: It looks like a `CipherInputStream` requires that the given `Cipher` be initialized already: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/javax/crypto/CipherInputStream.java?av=f#28.

However, a `CryptoInputStream` will actually initialize the given `CryptoCipher` for you using information that you give it, which includes the parameter spec: https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java#L488

So I'm glad you asked, because now it occurs to me that we're probably doing extra work by initializing the `CryptoCipher` before calling `CipherStreamSupplier#getInputStream` and `CipherStreamSupplier#getOutputStream`. I'll see if we can remove some cipher initialization from the diff here.

sjrand: Oh, whoops -- looking more closely, it's not just that a `CryptoInputStream` will initialize a `CryptoCipher` for you, it'll actually instantiate a new one. Which is pretty awkward, since I think that now we have two `CryptoCipher` instances: one that the `CryptoInputStream` creates, and another that we create in `SeekableCipher#initCipher`

sjrand: Do we want to just have `SeekableCipher#init` and `SeekableCipher#seek` return `void` then? That way they're setting up the `KeyMaterial`, the algorithm, and the `IvParameterSpec` for passing to the `CryptoInputStream`, but not creating a `CryptoCipher` that we wind up never using.

sjrand:
I don't think you guys are going to like the last commit, as it makes this change even more disruptive than it already was, but it turns out to be the logical continuation of the previous changes. As @uschi2000's question as to why a `CryptoInputStream` wants an `IvParameterSpec` prompted me to find out, the input/output streams in commons-crypto will create a `CryptoCipher` for you under the hood, whereas the javax streams expect you to give them an already initialized `Cipher`.

So now that the streams themselves take care of `CryptoCipher` creation, there's no reason for `initCipher` and `seek` in `SeekableCipher` to do the same -- thus the changes to the return types and method names.

sjrand:
Also, @ellisjoe and @uschi2000, I'm in a very different timezone from you guys (UTC+08:00), so if you think it'd be productive, I'd be down to stay up late and hop on a vtc to try to figure out how to proceed -- I'm guessing iteration speed over comments will likely not be great given the time gap.

ellisjoe:
@sjrand I've got a proposal for a less disruptive change that I'll put up today. Hopefully that helps move this along

sjrand:
Cool -- headed to bed soon, but I'll check it out tomorrow morning.

ellisjoe:
Found CtrCryptoInputStream which makes for a pretty low lift addition here: https://github.com/palantir/hadoop-crypto/pull/63

sjrand:
Had a Spark job write 14g into HDFS from a single task on a single executor and profiled the executor with yourkit. Spark code in both cases was just:

```
sc.parallelize(1 to (14*1024*1024)).map(k => Seq(k, org.apache.commons.lang.RandomStringUtils.random(1024, "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWxyZ0123456789")).mkString("|")).toDF().write.parquet("ehdfs://pal-hdfs/test")
```

* Latest develop spent about 112.5 seconds in FsCipherOutputStream#write:

image

Meanwhile, this PR only spent about 6.5 seconds:

image

sjrand:
And some read benchmarks. I wrote 4g of Parquet into HDFS and read it with this Spark code: 

```
spark.read.parquet("ehdfs://pal-hdfs/test-cache-me").cache().count()
```

Develop spent about 33 seconds in FsCipherInputStream#read:

image

Meanwhile, this PR spent about 4.3 seconds:

image

It looked from the snapshots like YourKit itself may have been altering the timings, so I also ran the same Spark code without YourKit, doing the timing in Scala instead (code stolen from the internet):

```
def time[T](f: => T) = {
val start = System.nanoTime
val rval = f
println("time: " + ((System.nanoTime - start) * 1e-9) + " s")
rval }
```

Results for 3 runs (new spark-shell for each run) each were:

* Develop: 118 seconds, 120 seconds, 107 seconds
* This PR: 54 seconds, 91 seconds, 61 seconds

The second run of this PR and the third run of develop are a little weird, but I think the numbers are pretty compelling nonetheless. 

It would also be interesting of course to run similar tests with `es3a`. I can look into that at some point when it's not the weekend.

sjrand:
@ellisjoe, what's the timeline for your version of incorporating the native crypto code? Given how useful this appears to be, I'd like for there to a strong sense of urgency in getting this over the finish line. Obviously I've done a terrible job of that by waiting 4 months to resume work on this, but I'd like to do better from here out.

I'm willing to make responding to feedback on this PR my top priority until it's done. Or, if we don't like this way, let me if there's anything I can do to speed things along with the alternate approach. 

Basically I don't want this to take much longer than it needs to just because we don't want to break APIs -- I'm aware of a couple projects using `ehdfs` soon/now, and don't want them to have a bad experience for avoidable reasons.

ellisjoe:
This PR merged https://github.com/palantir/hadoop-crypto/pull/63 and I'm planning to get it wired in to the EncryptedFileSystem this week.

ellisjoe:
wired up here: https://github.com/palantir/hadoop-crypto/pull/65

closing this since that PR uses AES-NI by default. let me know if there's any other concerns

sjrand:
Couple things:

* Need to make sure that perf of #65 is at least as good as perf of this. I imagine it should be, but need to check just in case. I can handle that part.
* Need to use OpenSSL for writes as well. What's the plan for doing that?

sjrand:
I'll open issues to track the above two things.

PR #36, opened 2017-01-10
move packages for v2

markelliot:


build.gradle
```diff
@@ -26,7 +26,7 @@ repositories {
 }
 
 allprojects {
-    group 'com.palantir.hadoop-crypto'
+    group 'com.palantir.hadoop-crypto2'
```
markelliot, line 29: probably end up having to mess with some stuff on bintray out of this

markelliot:
not sure if we actually have to move all the crypto stuff to a `.crypto2` or just the hadoop stuff

uschi2000:
I'd probably just move it all.

markelliot:
This PR moves everything but seek-io, which I intend to move to another project entirely (PR is in to our OSS management repo to get that provisioned.

ellisjoe (approved):

PR #37, opened 2017-01-10
Prevent even more negative seeking

markelliot:


crypto-core/src/test/java/com/palantir/crypto/cipher/NoNegativeSeeksTest.java
```diff
@@ -0,0 +1,55 @@
...
+    private static class DisallowNegativeSeeksSeekableInput implements SeekableInput {
+        private long pos = 0;
+
+        @Override
+        public void close() throws IOException {}
+
+        @Override
+        public void seek(long offset) throws IOException {
+            if (offset < pos) {
+                fail();
```
bavardage, line 38: fail message :P?

bavardage (approved):

crypto-core/src/test/java/com/palantir/crypto/cipher/NoNegativeSeeksTest.java
```diff
@@ -0,0 +1,55 @@
...
+import com.palantir.crypto.io.DecryptingSeekableInput;
+import com.palantir.seekio.SeekableInput;
+import java.io.IOException;
+import org.junit.Test;
+
+public final class NoNegativeSeeksTest {
+
+    @Test
+    public void testDecryptingSeekableInput_doesNotSeekNegatively() throws IOException {
+        for (int increment = 16; increment < 2048; increment += 16) {
```
uschi2000, line 18: does this only hold for `i*16` seeks? otherwise, why not test for all integers?

markelliot:
Can test all the integers, just picked the block size to make it go a bit faster (this test takes about 2s as written).

PR #39, opened 2017-01-29
Use OSS seek-io

ellisjoe:


build.gradle
```diff
@@ -52,6 +52,10 @@ subprojects {
 
     repositories {
         jcenter()
+        maven {
+            // TODO(jellis): figure out why seek-io and hadoop-crypto aren't mirrored to jcenter
```
schlosna: Seems like seek-io is still private, that could be why.

uschi2000:
Don't think that's the reason. Did you check the 'publish to jcenter'
button in bjntray for this project?

ellisjoe:
found the add to jcenter button. said it sent a message to jcenter...

ellisjoe:
@uschi2000 should be good now

versions.props
```diff
@@ -1,5 +1,6 @@
 com.fasterxml.jackson.*:jackson-* = 2.6.1
 com.google.guava:guava = 18.0
+com.palantir.seek-io:seek-io = 0.0.2
```
uschi2000: nit: would start with 0.1.0, but doesn't matter too much.

ellisjoe: done

markelliot (approved):

PR #44, opened 2017-02-08
manual wire serialization

ellisjoe:


ellisjoe:
@uschi2000 related to the serialization we talked about earlier

uschi2000:
in place of this, you're making an API class in the key service, ja?

ellisjoe:
correct. Serializable* will be replaced by api objects in the services

uschi2000:
cool. +1

PR #46, opened 2017-03-10
Delete key material when deleting files

sandorw:


sandorw:
@ellisjoe 

I chose to delete the key material first, since if you have to retry in the opposite order, the file has already been deleted and you can't tell if it was a dir or a file.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -98,6 +100,23 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+        FileStatus status;
+        try {
+            status = getFileStatus(path);
+        } catch (FileNotFoundException e) {
+            return false;
+        }
+
+        //Delete the key material if the path is a file
+        if (status.isFile()) {
+            keyStore.remove(path.toString());
```
ellisjoe: what happens if you've already deleted the keyMaterial?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -98,6 +100,23 @@ public boolean rename(Path src, Path dst) throws IOException {
         return renamed;
     }
 
+    @Override
+    public boolean delete(Path path, boolean recursive) throws IOException {
```
ellisjoe: could this just be
```
fs.delete(path, recursive);
keyStore.remove(path.toString());
```
That way the delete call takes care of figuring out if it's a directory or file.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
sandorw: Both key storage strategies swallow failures, assuming that the underlying FS returns false in the case of a nonexistent file, which seems to be the case at least for hdfs/s3.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
sandorw: I was trying to avoid deleting key material if a folder path is passed in for a recursive delete, but as noted in the other comment, it should be safe even if there's no key material.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -98,6 +98,15 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+        if (!fs.delete(path, recursive)) {
+            return false;
+        }
+        keyStore.remove(path.toString());
```
sandorw: If we want to handle retries here, in case the key store fails to remove the key material but the file was deleted, we'll have to do an existence check on the file first.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -266,4 +266,67 @@ public void testGetCipherAlgorithm_nonDefault() throws IOException, URISyntaxExc
...
+    @Test
+    public void testDelete_recursiveDeleteOnDir() throws IOException {
+        OutputStream os = efs.create(path);
+        os.write(0x00);
+        os.close();
+
+        Path folderPath = new Path(folder.getRoot().getAbsolutePath());
+        assertTrue(efs.delete(folderPath, true));
+
+        assertFalse(efs.exists(path));
```
ellisjoe: also assert that the key material is deleted

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -266,4 +266,67 @@ public void testGetCipherAlgorithm_nonDefault() throws IOException, URISyntaxExc
...
+        OutputStream os = efs.create(path);
+        os.write(0x00);
+        os.close();
+
+        Path folderPath = new Path(folder.getRoot().getAbsolutePath());
+        try {
+            efs.delete(folderPath, false);
+            fail();
+        } catch (IOException e) {
+            String expectedMessage = String.format("Directory %s is not empty", folderPath);
```
ellisjoe: inline

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -266,4 +266,67 @@ public void testGetCipherAlgorithm_nonDefault() throws IOException, URISyntaxExc
...
+
+        keyStore.remove(path.toString());
+        assertThat(keyStore.get(path.toString()), is(nullValue()));
+
+        assertTrue(efs.delete(path, false));
+
+        assertFalse(efs.exists(path));
+    }
+
+    private void testDeleteFile(boolean recursive) throws IOException {
```
ellisjoe: move this next to the two methods that call it

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -266,4 +266,67 @@ public void testGetCipherAlgorithm_nonDefault() throws IOException, URISyntaxExc
...
+            efs.delete(folderPath, false);
+            fail();
+        } catch (IOException e) {
+            String expectedMessage = String.format("Directory %s is not empty", folderPath);
+            assertThat(e.getMessage(), is(expectedMessage));
+        }
+    }
+
+    @Test
+    public void testDelete_keyMaterialAlreadyDeleted() throws IOException {
```
ellisjoe, line 300: fine to leave the test, but this should never happen except if someone manually removed the key material, correct?

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -266,4 +266,67 @@ public void testGetCipherAlgorithm_nonDefault() throws IOException, URISyntaxExc
...
+    @Test
+    public void testDelete_recusiveDeleteOnFile() throws IOException {
+        testDeleteFile(true);
+    }
+
+    @Test
+    public void testDelete_recursiveDeleteOnDir() throws IOException {
+        OutputStream os = efs.create(path);
+        os.write(0x00);
+        os.close();
```
ellisjoe: move this block into the before method (there's another one to move above in `testRename_successful`

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: can you leave a comment here about not being able to remove from the keystore all the time because we're not able to differentiate between file not found and the path being a directory?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -98,6 +98,18 @@ public boolean rename(Path src, Path dst) throws IOException {
         return renamed;
     }
 
+    @Override
+    public boolean delete(Path path, boolean recursive) throws IOException {
+    	if (fs.exists(path)) {
+    		boolean deleted = fs.delete(path, recursive);
```
ellisjoe: inline

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -98,6 +98,18 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+    		if (!deleted) {
+    			return false;
+    		}
+    	}
+        keyStore.remove(path.toString());
+        return true;
```
ellisjoe: how about this:
```
    public boolean delete(Path path, boolean recursive) throws IOException {
        // Existence check allows for unused keys to be cleaned up
        boolean exists = fs.exists(path);
        boolean deleted = fs.delete(path, recursive);

        if (deleted || !exists) {
            keyStore.remove(path.toString());
        }

        return deleted;
    }
```

sandorw: +1 - looks better than inlining to me

sandorw: Slight change needed at the end - return (deleted || !exists)

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
sandorw: Can't - the key material would only be deleted as a side effect of the recursive delete if you're using the file based strategy. Recursive deletes won't remove in-memory key material unless we want to add some logic to look at directory prefixes in the key storage map. The test uses the in-memory strategy.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
sandorw: Yep, made more sense with the previous impl where the key material was deleted first.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
sandorw: Nevermind, keep forgetting that non-existence means returning false :(

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -266,4 +266,57 @@ public void testGetCipherAlgorithm_nonDefault() throws IOException, URISyntaxExc
...
+
+        assertFalse(efs.exists(path));
+    }
+
+    @Test
+    public void testDelete_fileAlreadyDeleted() throws IOException {
+        delegateFs.delete(path, false);
+
+        assertFalse(efs.delete(path, false));
+        assertThat(keyStore.get(path.toString()), is(nullValue()));
```
ellisjoe: `assertThat(keyStore.get(path.toString()), is(keyMaterial));` above the `efs.delete` call.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: yep. we just rely on the delegate's answer.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
sandorw: Best I can do is assert that it's non-null - the only way in the non-mocked tests to get the key material is to directly read it from keyStore.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -98,25 +100,39 @@ public boolean rename(Path src, Path dst) throws IOException {
...
 
-        if (deleted || !exists) {
-            keyStore.remove(path.toString());
+    private boolean recursiveDelete(Path path) throws IOException {
+        //List all files under the path recursively and delete their key material
+        RemoteIterator<LocatedFileStatus> statuses = fs.listFiles(path, true);
+        while (statuses.hasNext()) {
+            LocatedFileStatus status = statuses.next();
+            if (status.isFile()) {
+                Path filePath = Path.getPathWithoutSchemeAndAuthority(status.getPath());
```
sandorw: @ellisjoe double check this - I believe this is fine since we're already wrapping the correct file system at this point. Existing tests need the scheme to be missing.

sandorw:
Switched back to deleting the key material first to handle the recursive case properly. In-memory key storage should be updated properly (and any other extension) when a recursive delete happens.

ellisjoe:
what do you mean the key storage strategy needs to be updated properly?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: why do we need this here? we just use toString everywhere else?

sandorw: fs.listFiles seems to return paths that include the scheme. File based key storage is likely fine with this, but the in memory storage strategy fails if the keys are suddenly prepended with file:/, as was the case in the existing tests.

ellisjoe (approved):

ellisjoe (approved):

ellisjoe:
you have write perms now. need to repush without the fork

PR #52, opened 2017-04-04
move to java8 where possible

ellisjoe:


crypto-keys/src/main/java/com/palantir/crypto2/keys/KeyPairs.java
```diff
@@ -41,15 +41,15 @@ public static KeyPair fromStrings(String privateKeyString, String publicKeyStrin
         try {
             KeyFactory keyFactory = KeyFactory.getInstance(algorithm);
             // Private key is only required for decryption, can be null
-            Optional<PrivateKey> privateKey = Optional.absent();
+            Optional<PrivateKey> privateKey = Optional.empty();
             if (privateKeyString != null) {
-                KeySpec privateKs = new PKCS8EncodedKeySpec(BaseEncoding.base64().decode(privateKeyString));
+                KeySpec privateKs = new PKCS8EncodedKeySpec(Base64.getDecoder().decode(privateKeyString));
```
uschi2000, line 46: maybe check these are equivalent

uschi2000 (approved):

crypto-keys/src/main/java/com/palantir/crypto2/keys/KeyPairs.java
ellisjoe: verified:
```
    @Test
    public void testJavaAndGuavaBase64Equivalent() {
        byte[] bytes = "bytes".getBytes(StandardCharsets.UTF_8);

        String guavaEncoded = BaseEncoding.base64().encode(bytes);
        byte[] javaDecoded = Base64.getDecoder().decode(guavaEncoded);

        assertThat(javaDecoded, is(bytes));

        String javaEncoded = Base64.getEncoder().encodeToString(bytes);
        byte[] guavaDecoded = BaseEncoding.base64().decode(javaEncoded);

        assertThat(guavaDecoded, is(bytes));
    }
```

PR #53, opened 2017-04-06
Allow overriding exists (which FilterFileSystem doesn't override)

robert3005:


ellisjoe (approved):

sjrand:
Purely out of curiosity, what motivated this change?

robert3005:
We want to use FileSystem#exists but since we wrap every filesystem in EncryptedFileSystem we are limited to default impl which uses getFileStatus. It's a lot simpler to implement exists than getFileStatus so would like to avoid unnecessary indirection.

sjrand:
Makes sense, thanks

PR #60, opened 2017-04-14
replace ByteArraySeekableInput with InMemorySeekableDataInput

ellisjoe:


ellisjoe:
@markelliot fixes https://github.com/palantir/hadoop-crypto/issues/40

PR #61, opened 2017-04-20
decryption benchmarks

ellisjoe:


hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/benchmarks/BenchmarkReadWrite.java
```diff
@@ -0,0 +1,153 @@
...
+        IvParameterSpec iv = new IvParameterSpec(state.keyMaterial.getIv());
+        return new CryptoInputStream(AesCtrCipher.ALGORITHM, state.properties, is, key, iv);
+    }
+
+    private static void readFully(SeekableInput input) throws IOException {
+        ByteStreams.toByteArray(new DefaultSeekableInputStream(input));
+    }
+
+    private static void readFully(InputStream input) throws IOException {
+        ByteStreams.toByteArray(input);
```
schlosna: Return byte array from here and also return from benchmark callers

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/benchmarks/BenchmarkReadWrite.java
```diff
@@ -0,0 +1,153 @@
...
+    private static CryptoInputStream getApacheCryptoStream(MyState state, SeekableInput input)
+            throws IOException {
+        DefaultSeekableInputStream is = new DefaultSeekableInputStream(input);
+        SecretKey key = state.keyMaterial.getSecretKey();
+        IvParameterSpec iv = new IvParameterSpec(state.keyMaterial.getIv());
+        return new CryptoInputStream(AesCtrCipher.ALGORITHM, state.properties, is, key, iv);
+    }
+
+    private static void readFully(SeekableInput input) throws IOException {
+        ByteStreams.toByteArray(new DefaultSeekableInputStream(input));
```
schlosna: Return byte array from here and also return from benchmark callers

ellisjoe: done

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/benchmarks/BenchmarkReadWrite.java
ellisjoe: done

crypto-core/src/test/java/com/palantir/crypto2/benchmarks/BenchmarkReadWrite.java
```diff
@@ -0,0 +1,156 @@
...
+
+            keyMaterial = AesCtrCipher.generateKeyMaterial();
+        }
+
+    }
+
+    @Benchmark
+    @Threads(value = 1)
+    @Warmup(iterations = 0)
+    @BenchmarkMode(value = Mode.SingleShotTime)
```
schlosna, line 95: Curious why we're doing single shot and no warmups? I think we probably need to do some warmup, especially as each path likely has some expensive initialization (e.g. JCE services, JNI, etc.)

crypto-core/src/test/java/com/palantir/crypto2/benchmarks/SeekableInputChannel.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public long getPos() throws IOException {
+        return channel.position();
+    }
+
+    @Override
+    public int read(byte[] bytes, int offset, int length) throws IOException {
+        return channel.read(ByteBuffer.wrap(bytes, offset, length));
```
schlosna, line 44: If we want to support this, we probably want to better support ByteBuffer as destination in SeekableInput with a `read(ByteBuffer destination)` per 
https://docs.oracle.com/javase/8/docs/api/java/nio/channels/SeekableByteChannel.html#read-java.nio.ByteBuffer- and 
https://docs.oracle.com/javase/8/docs/api/java/nio/channels/ReadableByteChannel.html#read-java.nio.ByteBuffer-

schlosna (commented):
I'll clone and test in a bit, but curious if you could include some baseline numbers for comparison.

crypto-core/src/test/java/com/palantir/crypto2/benchmarks/BenchmarkReadWrite.java
ellisjoe: this is very much me just playing around to get a feel for how the JCE and Apache libraries compare. happy to take any and all advice here as far as configuring the tests

crypto-core/src/test/java/com/palantir/crypto2/benchmarks/SeekableInputChannel.java
ellisjoe: don't think this is necessary. wanted to test if i could use channels as nonblocking io to get around the fact that currently `read rate = read from s3/disk + decryption` rather than `read rate = min(read from s3/disk, decryption)`

PR #63, opened 2017-04-25
AES-NI support through apache commons crypto

ellisjoe:


markelliot (approved):

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -0,0 +1,114 @@
...
+import com.palantir.crypto2.keys.KeyMaterial;
+import com.palantir.seekio.SeekableInput;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Properties;
+import org.apache.commons.crypto.stream.CtrCryptoInputStream;
+import org.apache.commons.crypto.stream.input.Input;
+import org.apache.commons.crypto.utils.Utils;
+
+public final class ApacheCtrDecryptingSeekableInput extends CtrCryptoInputStream implements SeekableInput {
```
uschi2000: docs

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -0,0 +1,114 @@
...
+    private ApacheCtrDecryptingSeekableInput(SeekableInput input, KeyMaterial keyMaterial) throws IOException {
+        super(new InputAdapter(input), Utils.getCipherInstance(ALGORITHM, new Properties()), BUFFER_SIZE,
+                keyMaterial.getSecretKey().getEncoded(), keyMaterial.getIv());
+    }
+
+    public static ApacheCtrDecryptingSeekableInput create(SeekableInput input, KeyMaterial keyMaterial) {
+        try {
+            return new ApacheCtrDecryptingSeekableInput(input, keyMaterial);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
uschi2000, line 50: `throw new RuntimeException("Failed to...", e);`

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -0,0 +1,114 @@
...
+            input.seek(position);
+            return input.read(buffer, offset, length);
+        }
+
+        @Override
+        public int read(ByteBuffer dst) throws IOException {
+            byte[] bytes = new byte[dst.capacity() - dst.position()];
+            int read = input.read(bytes, 0, bytes.length);
+
+            if (read != -1) {
```
uschi2000, line 100: does this handle the `read == 0` case correctly?

crypto-core/src/test/java/com/palantir/crypto2/io/DecryptingSeekableInputTest.java
```diff
@@ -33,6 +33,8 @@
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Random;
+import java.util.function.BiFunction;
+import javafx.util.Pair;
```
schlosna: probably don't want this specific Pair from javafx

crypto-core/src/test/java/com/palantir/crypto2/io/DecryptingSeekableInputTest.java
```diff
@@ -62,18 +64,37 @@ public static void beforeClass() throws IOException {
...
+        return ImmutableList.of(
+                new Pair<>(AesCtrCipher.ALGORITHM, DecryptingSeekableInputTest::apacheStream),
+                new Pair<>(AesCtrCipher.ALGORITHM, DecryptingSeekableInputTest::jceStream),
+                new Pair<>(AesCbcCipher.ALGORITHM, DecryptingSeekableInputTest::jceStream));
     }
 
-    public DecryptingSeekableInputTest(String cipher) {
+    private static SeekableInput apacheStream(SeekableCipher cipher, SeekableInput input) {
+        if (cipher instanceof AesCtrCipher) {
+            return ApacheCtrDecryptingSeekableInput.create(input, cipher.getKeyMaterial());
```
schlosna, line 75: how are we verifying that we're using the OpenSSL AES-NI enabled version?

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
ellisjoe: done

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
ellisjoe: we don't do this anywhere else

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
ellisjoe: is there something specific you think might now work? call `put` with when `read == 0` is a no-op.

crypto-core/src/test/java/com/palantir/crypto2/io/DecryptingSeekableInputTest.java
ellisjoe: done

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
uschi2000: I think we should

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
uschi2000: nothing specific I had in mind, just double-checking.

crypto-core/src/test/java/com/palantir/crypto2/io/DecryptingSeekableInputTest.java
ellisjoe: modified so this class can only be constructed with OpenSSL and won't fall back to JCE. also added an `isSupported` method that checks if OpenSSL was loaded.

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
ellisjoe: Can make the change everywhere as a separate PR then.

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
ellisjoe: yeah should be good to go

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -0,0 +1,134 @@
...
+ * A {@link SeekableInput} that decrypts AES/CTR encrypted SeekableInputs using the given {@link KeyMaterial}. This
+ * implementation uses Apache's {@link CtrCryptoInputStream} which uses OpenSSL and supports AES-NI.
+ */
+public final class ApacheCtrDecryptingSeekableInput extends CtrCryptoInputStream implements SeekableInput {
+
+    private static final String ALGORITHM = "AES/CTR/NoPadding";
+    private static final int BUFFER_SIZE = 8192;
+    private static final Properties PROPS;
+
+    static {
```
markelliot: generally prefer you make this a static method and then call:

    private static final Properties PROPS = initializeProps();

ellisjoe: done

sjrand (commented):
Should this also change `FsCipherInputStream` to use an `ApacheCtrDecryptingSeekableInput`?

Also curious as to whether there's a plan for writes as well as reads, but not blocking.

ellisjoe:
@sjrand will wire this into FsCipherInputStream in a separate PR. can add write support then as well.

PR #64, opened 2017-04-26
fix test when running on mac

ellisjoe:


gradle/openssl.gradle
```diff
@@ -0,0 +1,7 @@
+test { 
+    if (org.gradle.internal.os.OperatingSystem.current().isMacOsX()) {
+        def opensslDir = "/usr/local/opt/openssl/lib"
```
schlosna: Can we append this to `LD_LIBRARY_PATH` so that if one doesn't have openssl in `/usr/local/opt/openssl/lib` we can still make it work?

ellisjoe: done

schlosna (commented):
Thanks Joe, :+1: LGTM

PR #65, opened 2017-05-01
Hook in Apache DecryptingSeekableInput

ellisjoe:


crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -54,7 +54,7 @@ public static ApacheCtrDecryptingSeekableInput create(SeekableInput input, KeyMa
     /**
      * Returns true if the OpenSSL is able to be loaded.
      */
-    public boolean isSupported() {
+    public static boolean isSupported() {
```
uschi2000: this name seems misleading. This class always works correctly, but only sometimes uses the native code, correct?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/cipher/FsCipherInputStream.java
```diff
@@ -31,11 +33,21 @@
 
     private final DefaultSeekableInputStream delegate;
 
+    /**
+     * @deprecated use {@link FsCipherInputStream#FsCipherInputStream(FSDataInputStream, KeyMaterial, String)} instead.
```
uschi2000, line 38: class-level docs are stale

uschi2000: (in particular, should spec what "algorithm" is..

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
uschi2000: in particular, what happens if I instantiate this class when `isSupported() == false`?

ellisjoe: it'll fail when you try to instantiate it. the key part being the fact that I explicitly pass in properties that force it to use OpenSSL and fail otherwise: https://github.com/palantir/hadoop-crypto/blob/develop/crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java#L129-L133

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/cipher/FsCipherInputStream.java
ellisjoe: done

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
uschi2000: ok. should probably document that behavior a little more careful.

uschi2000:
```
/** Creates a new ... . This constructor is expected to succeed if and only if {@link #isSupported} returns true. Callers should check {@link #isSupported} before calling this constructor. */
public static ApacheCtrDecryptingSeekableInput create(..) { ... }
```

crypto-core/src/main/java/com/palantir/crypto2/io/DecryptingSeekableInputFactory.java
```diff
@@ -0,0 +1,40 @@
...
+import com.palantir.crypto2.keys.KeyMaterial;
+import com.palantir.seekio.SeekableInput;
+
+public final class DecryptingSeekableInputFactory {
+
+    private DecryptingSeekableInputFactory() {}
+
+    /**
+     * Returns a {@link SeekableInput} that decrypts the given SeekableInput using the given {@link KeyMaterial} and
+     * cipher {@code algorithm}.
```
uschi2000: should mention that sweet on OpenSSl availability

uschi2000 (approved):
looks good, but I'd prefer to be a bit more explicit on the documentation side. the behavior is pretty non-obvious for anyone not deep in the weeds here.

ellisjoe:
closing in favor of https://github.com/palantir/hadoop-crypto/pull/70

PR #70, opened 2017-05-08
CryptoStreamFactories

ellisjoe:


ellisjoe:
alright this cleans up and combines the other two PRs into one Factory. Only other pieces are wiring in the Factory usages and deprecated FsCipherOutputStream, which needlessly wraps CipherOutputStream.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -89,7 +89,9 @@ public FSDataOutputStream create(Path path, FsPermission permission,
         SeekableCipher cipher = SeekableCipherFactory.getCipher(cipherAlgorithm);
 
         // Ensure we can open the stream before storing keys that would be irrelevant
-        FSDataOutputStream os = new FSDataOutputStream(new FsCipherOutputStream(encryptedStream, cipher), null);
+        OutputStream encryptedOs =
+                CryptoStreamFactory.encrypt(encryptedStream, cipher.getKeyMaterial(), cipherAlgorithm);
+        FSDataOutputStream os = new FSDataOutputStream(encryptedOs, null);
```
schlosna: Do we not want to track `org.apache.hadoop.fs.FileSystem.Statistics` -- should we be passing the protected `statistics` as final arg rather than null?

ellisjoe: done

schlosna (approved):
👍

Issue #27, opened 2016-12-07
Fix StandaloneEncryptedFileSystem#copyFromLocalFile


ellisjoe:
It currently does not encrypt to file when writing due to https://issues.apache.org/jira/browse/HADOOP-13870

uschi2000:
oops

PR #72, opened 2017-05-10
implement flavor of FileSystem#create used by distcp for intermediate files

### LINKED ISSUE #71: cannot distcp into ehdfs
Hadoop's distcp utility first writes output to a temporary directory in the target filesystem: https://github.com/apache/hadoop/blob/branch-2.8.0/hadoop-tools/hadoop-distcp/src/main/java/org/apache/hadoop/tools/mapred/RetriableFileCopyCommand.java#L174. This intermediate data is later copied to the destination `Path`.

Currently, `EncryptedFileSystem` does not implement the version of `create()` that is called in the linked code. Therefore the attempt to write intermediate data to `ehdfs` fails, and instead, the mapper falls back to writing to `hdfs`. Therefore the data is not encrypted, and there is no symmetric key.

The lack of a symmetric key (correctly) causes `EncryptedFileSystem#rename` to fail when the intermediate data is promoted to the target path: `java.lang.RuntimeException: java.io.FileNotFoundException: File does not exist: /.distcp.tmp.attempt_local913233795_0001_m_000000_0.keymaterial`.

sjrand:
fixes #71


ellisjoe:
happy to merge this once the conflicts are fixed. think the long term fix is to stop using `FilterFileSystem` and implement `FileSystem` directly. saw similar problems here: https://github.com/palantir/hadoop-crypto/issues/27

sjrand:
Conflicts should be fixed now.

ellisjoe (approved):

sjrand:
thanks!

ellisjoe:
cut 2.1.1 with this

PR #73, opened 2017-05-25
Add async methods to KeyStorageStrategy

j-baker:
At the moment, the KeyStorageStrategy is blocking - it'd be really nice if there was an async version people could use.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 buildscript {
     repositories {
-        jcenter()
         maven { url  "http://palantir.bintray.com/releases" }
+        jcenter()
     }
 
     dependencies {
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 buildscript {
     repositories {
         jcenter()
+        maven { url  "http://palantir.bintray.com/releases" }
     }
 
     dependencies {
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -1,5 +1,6 @@
 buildscript {
     repositories {
+        maven { url  "http://palantir.bintray.com/releases" }
```
markelliot, line 3: what is this used for?

crypto-keys/src/main/java/com/palantir/crypto2/keys/ChainedKeyStorageStrategy.java
```diff
@@ -65,6 +69,34 @@ public KeyMaterial get(String fileKey) {
     }
 
     @Override
+    public CompletableFuture<KeyMaterial> getAsync(String fileKey) {
```
uschi2000, line 71: wow, this is complicated.

crypto-keys/src/main/java/com/palantir/crypto2/keys/KeyStorageStrategy.java
```diff
@@ -32,6 +34,18 @@
...
 
     /**
+     * Retrieves the {@link KeyMaterial} for a file with the given {@code fileKey},
+     * returning a {@link CompletableFuture}. Consumers should expect this to execute
+     * in a nonblocking fashion.
+     *
+     * By default uses the common fork-join pool - implementers will likely want to
+     * change this behaviour in non-trivial cases.
+     */
+    default CompletableFuture<KeyMaterial> getAsync(String fileKey) {
```
markelliot, line 44: likely better to have a separate interface called `AsyncKeyStorageStrategy`

PR #74, opened 2017-05-25
Add AsyncKeyStorageStrategy

ellisjoe:


ellisjoe:
alternate proposal to https://github.com/palantir/hadoop-crypto/pull/73

crypto-keys/src/main/java/com/palantir/crypto2/keys/DefaultAsyncKeyStorageStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+            new ThreadFactoryBuilder().setNameFormat("async-key-storage-strategy-%s").build());
+
+    private final KeyStorageStrategy keys;
+    private final ExecutorService executor;
+
+    public DefaultAsyncKeyStorageStrategy(KeyStorageStrategy keys) {
+        this(keys, DEFAULT_EXECUTOR);
+    }
+
+    @VisibleForTesting
```
uschi2000: think I would make this the default: make it explicit that you need to provide an executor service

crypto-keys/src/main/java/com/palantir/crypto2/keys/AsyncKeyStorageStrategy.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * The asynchronous strategy used to store the per file {@link KeyMaterial} used for encryption operations.
+ */
+public interface AsyncKeyStorageStrategy {
+
+    /**
+     * Stores the {@link KeyMaterial} for a file with the given {@code fileKey}.
```
uschi2000: would just delegate to the KeyStoreStrategy docs, no need to repeat.

crypto-keys/src/main/java/com/palantir/crypto2/keys/DefaultAsyncKeyStorageStrategy.java
ellisjoe: don't think I want consumers to provide an executor service. If I force that then this class becomes pretty pointless

crypto-keys/src/main/java/com/palantir/crypto2/keys/AsyncKeyStorageStrategy.java
ellisjoe: done.

crypto-keys/src/main/java/com/palantir/crypto2/keys/DefaultAsyncKeyStorageStrategy.java
markelliot: disagree, we should be DIing the ExecutorService so it can be managed appropriately

ellisjoe: done

j-baker (approved):

crypto-keys/src/test/java/com/palantir/crypto2/keys/DefaultAsyncKeyStorageStrategyTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @Before
+    public void before() {
+        keyMaterial = mock(KeyMaterial.class);
+        delegate = mock(KeyStorageStrategy.class);
+
+        keys = new DefaultAsyncKeyStorageStrategy(delegate, executor);
+    }
+
+    @Test
+    public void testPut() {
```
uschi2000, line 48: would be good to have one example for handling errors too.

uschi2000 (approved):

PR #75, opened 2017-05-25
test that all methods decrypt apache encrypted streams

ellisjoe:


ellisjoe:
merging since this is almost all additional testing. `ApacheCiphers.forceOpenSsl` is the only public api addition.

PR #76, opened 2017-05-26
have EncryptedFileSystem implement FileSystem directly

ellisjoe:


ellisjoe:
should fix https://github.com/palantir/hadoop-crypto/issues/27 and deals with https://github.com/palantir/hadoop-crypto/pull/72

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -359,4 +361,61 @@ public void testDelete_fileAlreadyDeleted() throws IOException {
...
+    public void testExists() throws IOException {
+        mockedEfs.exists(path);
+
+        verify(mockFs).exists(path);
+    }
+
+    @Test
+    public void testListStatus() throws IOException {
+        mockedEfs.listStatus(path);
+
```
uschi2000, line 378: just remove these lines? (unless you can give me a compelling reason for having them :) )

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -359,4 +361,61 @@ public void testDelete_fileAlreadyDeleted() throws IOException {
...
+        verify(mockFs).getUri();
+    }
+
+    @Test
+    public void testAppend() {
+        assertThatExceptionOfType(UnsupportedOperationException.class)
+                .isThrownBy(() -> mockedEfs.append(path, 0, null))
+                .withMessage("appending to encrypted files is not supported");
+    }
+
```
uschi2000, line 423: worth adding a test that protects against https://issues.apache.org/jira/browse/HADOOP-13870 ?

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
ellisjoe: it's a bit fanatic but I like grouping test methods like:
```
setup (initialization, mocking, etc)

run test methods

assertions
```

can remove since these are so short but it's how I've done it everywhere: https://github.com/palantir/hadoop-crypto/blob/develop/hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/PathConvertingFileSystemTest.java

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
ellisjoe: It's in `StandaloneEncryptedFileSystem` but can add it here as well

ellisjoe: done

uschi2000 (approved):

PR #77, opened 2017-05-26
use given key when constructing encrypting cipher

ellisjoe:


crypto-core/src/test/java/com/palantir/crypto2/io/CryptoStreamFactoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import org.junit.Test;
+
+public final class CryptoStreamFactoryTest {
+
+    @Test
+    public void testEncryptDecryptJce() throws IOException {
+        System.setProperty(CryptoStreamFactory.FORCE_JCE, "true");
```
schlosna: Probably want to remove this in an @After

ellisjoe: done

ellisjoe: just stuck it at the end of the test since it should only apply to this specific test at the moment

crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
```diff
@@ -80,8 +81,9 @@ private static OutputStream createApacheEncryptedStream(OutputStream output, Key
         return new CtrCryptoOutputStream(PROPS, output, secretKey.getEncoded(), iv);
     }
 
-    private static OutputStream createDefaultEncryptedStream(OutputStream output, String algorithm) {
-        SeekableCipher cipher = SeekableCipherFactory.getCipher(algorithm);
+    private static OutputStream createDefaultEncryptedStream(OutputStream output, KeyMaterial keyMaterial,
```
robert3005, line 95: Nit: would push all arguments to next line

robert3005 (approved):

crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
```diff
@@ -80,8 +81,9 @@ private static OutputStream createApacheEncryptedStream(OutputStream output, Key
...
+            String algorithm) {
+        SeekableCipher cipher = SeekableCipherFactory.getCipher(algorithm, keyMaterial);
```
uschi2000, line 97: the SeekableCipherFactory is dangerous. would either remove the method that does generates key-material, or rename it to getCipherWithGeneratedKeyMaterial, or make the key material an `Optional<>` parameter.

crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
```diff
@@ -44,7 +45,7 @@ private CryptoStreamFactory() {}
      * cipher {@code algorithm}. When OpenSSL is available an implementation that uses AES-NI will be returned.
      */
     public static SeekableInput decrypt(SeekableInput encryptedInput, KeyMaterial keyMaterial, String algorithm) {
-        if (!algorithm.equals(AES_ALGORITHM)) {
+        if (!algorithm.equals(AES_ALGORITHM) || Boolean.parseBoolean(System.getProperty(FORCE_JCE))) {
```
uschi2000: this seems like an orthogonal change? prefer 2 PRs, one to fix the bug and one to allow overriding JCE

uschi2000: oh, hang on. the FORCE_JCE thing is for testing only? that's a bit invasive, IMO.

uschi2000: oh, hang on. the FORCE_JCE thing is for testing only? that's a bit invasive, IMO.

uschi2000 (approved):

PR #79, opened 2017-05-26
deprecate dangerous factory method

ellisjoe:


crypto-core/src/main/java/com/palantir/crypto2/cipher/SeekableCipherFactory.java
```diff
@@ -26,6 +26,26 @@
 
     private SeekableCipherFactory() {}
 
+    public static KeyMaterial generateKeyMaterial(String cipherAlgorithm) {
```
uschi2000: docs

uschi2000 (approved):
lgtm modulo one docs nit

crypto-core/src/main/java/com/palantir/crypto2/cipher/SeekableCipherFactory.java
ellisjoe: done

PR #83, opened 2017-06-01
use assertj and remove hamcrest

ellisjoe:


ellisjoe:
will have PRs for the other projects after this merges

uschi2000 (approved):
didn't read every line

PR #87, opened 2017-07-29
Remove SunJCE hardcode

glentakahashi:
Removes the hardcode on provider so that it can work on systems without SunJCE, such as IBM's AIX.

### DIFF
--- a/crypto-core/src/main/java/com/palantir/crypto2/cipher/AesCbcCipher.java
+++ b/crypto-core/src/main/java/com/palantir/crypto2/cipher/AesCbcCipher.java
@@ -23,7 +23,6 @@ import com.palantir.crypto2.keys.serialization.KeyMaterials;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
-import java.security.NoSuchProviderException;
 import javax.crypto.Cipher;
 import javax.crypto.NoSuchPaddingException;
 import javax.crypto.SecretKey;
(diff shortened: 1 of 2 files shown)


uschi2000:
@mphillips

ellisjoe:
@glentakahashi which provider are you using? would be good to test against it and explicitly allow it here.

glentakahashi:
Unfortunately, IBMJCE
Not sure how easy it is to get + test that provider on a non-ibm system

uschi2000:
I can't judge whether this is safe to do, deferring to MattP.

uschi2000:
I can't judge whether this is safe to do, deferring to MatP.

ghost:
This looks safe to me :+1:

ellisjoe:
Think the main concern is around correctness. Not sure there's a way to test this besides running the tests on an IBM jvm which probably requires us to start using docker

ghost:
From Java 7, JCE providers are required to provide implementations of specific ciphers but unfortunately `AES/CTR/NoPadding` is not one of them. (See here: [6](http://docs.oracle.com/javase/6/docs/technotes/guides/security/StandardNames.html#impl), [7](http://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#impl), [8](http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#impl)).

If IBM's JCE doesn't support it, you could deploy with the Bouncy Castle JCE provider and change your 
JRE java.security policy file.

glentakahashi:
I think that would be a more reasonable error to have though. If the code requires specifically NoPadding, and the JCE you have installed doesn't support it, throwing an error and exiting is probably sufficient. It would be on the user to then install a compatible JCE. (FWIW IBM's JCE does support it afaik)

ellisjoe:
@glentakahashi have a PR that verifies things work on an ibm jvm: https://github.com/palantir/hadoop-crypto/pull/89

ellisjoe:
closing since the other PR merged

PR #88, opened 2017-08-02
Newer version of seek-io

j-baker:
An internal project depends on hadoop-crypto and when you depend on it, it generates dependency conflicts because it needs a newer version which requires manual resolution. Seems reasonable to pick up minor version bumps.


ash211:
Anything you're particularly interested in from the version bump?

j-baker:
Updated the title!

ash211 (approved):

ellisjoe (approved):

PR #89, opened 2017-08-15
verify ibm jvm compatability

ellisjoe:


glentakahashi:
👍

build.gradle
```diff
@@ -1,7 +1,7 @@
 buildscript {
     repositories {
-        jcenter()
         maven { url  "http://palantir.bintray.com/releases" }
+        jcenter()
```
uschi2000, line 4: critical change.

crypto-core/src/main/java/com/palantir/crypto2/cipher/Ciphers.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    @VisibleForTesting
+    static String getProvider(List<String> providers) {
+        for (String provider : providers) {
+            if (Security.getProvider(provider) != null) {
+                return provider;
+            }
+        }
+        throw new IllegalStateException(
+                String.format("None of the requested JCE providers are acceptable: %s", providers));
```
uschi2000: isn 't this: "None of the acceptable JCE providers are available" ?

Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM ibmjava:8-sdk
```
uschi2000, line 1: does this thing not contain SunJCE? (have you verified?)

crypto-core/src/main/java/com/palantir/crypto2/cipher/Ciphers.java
ellisjoe: done

build.gradle
ellisjoe: sadly it is

Dockerfile
ellisjoe: removing "IBMJCE" from the list of acceptable providers gives me `java.lang.IllegalStateException: None of the acceptable JCE providers are acceptable: [SunJCE]` when running the tests in this container.

uschi2000:
Ship it

PR #90, opened 2017-09-13
Delegate down default replication

pwoody:
@ellisjoe @uschi2000 @sjrand


sjrand:
Other FileSystem methods to probably delegate (maybe in a separate PR since this is more important):

* getFileBlockLocations
* getDefaultBlockSize

ellisjoe (approved):

markelliot:
Can we push for an abstract class DelegatingFileSystem implementation that overrides _every_ method and then extend that for EncryptedFileSystem? Concerned there are other things we're missing.

ellisjoe:
yeah. probably going to extend `FilterFileSystem` and override as necessary (eg: copyFromLocal)

PR #91, opened 2017-09-13
test delegation

ellisjoe:


pwoody:
@ellisjoe probably worth putting some integration tests for DFS on the docket.

PR #94, opened 2017-09-25
use delegating filesystem

ellisjoe:


hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/DelegatingFileSystem.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.hadoop.fs.FilterFileSystem;
+import org.apache.hadoop.fs.Path;
+
+/**
+ * Equivalent to {@link FilterFileSystem} but invokes {@link #create} and {@link #open} on this class when calling
+ * {@link #copyFromLocalFile} and {@link #copyToLocalFile}.
+ * <p>
+ * Solves: https://issues.apache.org/jira/browse/HADOOP-13870
+ */
+public abstract class DelegatingFileSystem extends FilterFileSystem {
```
uschi2000, line 32: should/could we not use FilterFileSystem at all?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/DelegatingFileSystem.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public DelegatingFileSystem(FileSystem delegate) {
+        super(delegate);
+        try {
+            super.initialize(delegate.getUri(), delegate.getConf());
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
uschi2000: "Failed to initialize delegate FileSystem"

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/DelegatingFileSystem.java
ellisjoe: we'd end up reimplementing the whole thing since we do want to delegate every other method

uschi2000: yeah, I know. just wondering how stable that FFS is over different Hadoop versions? do we rely on unsafe subclassing in a way that would bite us?

ellisjoe: think since it's just delegating we don't need to worry too much.

sjrand (approved):

ellisjoe:
@uschi2000 any other concerns?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/DelegatingFileSystem.java
```diff
@@ -0,0 +1,83 @@
...
+    public DelegatingFileSystem(FileSystem delegate) {
+        super(delegate);
+        try {
+            super.initialize(delegate.getUri(), delegate.getConf());
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to initialize the delegating filesystem", e);
+        }
+    }
+
+    public void copyFromLocalFile(Path src, Path dst) throws IOException {
```
uschi2000: no override? should add javadoc in that case

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/DelegatingFileSystem.java
```diff
@@ -0,0 +1,83 @@
...
+        copyFromLocalFile(false, src, dst);
```
uschi2000: consider calling the 4arg one directly

uschi2000 (approved):

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/DelegatingFileSystem.java
ellisjoe: done

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/DelegatingFileSystem.java
ellisjoe: prefer this so the default parameters are specified in exactly one method. this also matches the impl in `FileSystem`

PR #95, opened 2017-10-09
authenticated encryption disclaimer

ellisjoe:


cbarcenas:
What about something like:

> Neither supported AES mode is [authenticated](https://en.wikipedia.org/wiki/Authenticated_encryption). Authentication should be performed by consumers of this library via an external cryptographic mechanism such as Encrypt-then-MAC. 
> Failure to properly authenticate ciphertext breaks security in some scenarios where an attacker can manipulate ciphertext inputs.

ellisjoe:
that sounds fine to me. @uschi2000 sounds good to you?

uschi2000:
Yep.

-------- Original Message --------

PR #96, opened 2017-10-25
When we fail due to multiple strategies failing, add the error as suppressed exceptions

j-baker:


ellisjoe (approved):

crypto-keys/src/main/java/com/palantir/crypto2/keys/ChainedKeyStorageStrategy.java
```diff
@@ -53,17 +54,21 @@ public void put(String fileKey, KeyMaterial keyMaterial) {
...
+                suppressedExceptions.add(e);
                 logger.info("Failed to get key material using {}", strategy.getClass().getCanonicalName(), e);
             }
         }
-        throw new RuntimeException(String.format(
+        RuntimeException toThrow = new RuntimeException(String.format(
                 "Unable to get key material for '%s' using any of the provided strategies: %s",
                 fileKey,
                 Collections2.transform(strategies, s -> s.getClass().getCanonicalName())));
+        suppressedExceptions.stream().forEach(toThrow::addSuppressed);
```
ellisjoe, line 70: will remove `.stream()` once I switch to error-prone

PR #97, opened 2017-10-25
update baseline

ellisjoe:


crypto-core/src/main/java/com/palantir/crypto2/cipher/CipherStreamSupplier.java
```diff
@@ -1,17 +1,5 @@
 /*
- * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
```
ash211: do you intentionally want to remove Apache license?

ellisjoe: fixed. thanks

circle.yml
```diff
@@ -9,15 +9,15 @@ machine:
 
 dependencies:
   override:
-    - ./gradlew --parallel resolveConfigurations
+    - ./gradlew --parallel --no-daemon resolveConfigurations
```
uschi2000: add comment: `# Cannot use daemon mode since standard build and verify-ibmjvm.sh build use different plugins and classpaths, see build.gradle`

uschi2000 (approved):

PR #92, opened 2017-09-13
override getDefaultBlockSize and getFileBlockLocations

sjrand:

PR #99, opened 2017-11-23
Delegate getfileblocklocations with path

robert3005:


hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/DelegatingFileSystem.java
```diff
@@ -81,4 +84,9 @@ public void copyToLocalFile(boolean delSrc, Path src, Path dst, boolean useRawLo
                 : getLocal(conf);
         FileUtil.copy(this, src, local, dst, delSrc, conf);
     }
+
+    @Override
+    public BlockLocation[] getFileBlockLocations(Path path, long start, long len) throws IOException {
```
ash211, line 89: So did we lose https://github.com/palantir/hadoop-crypto/pull/92 when moving to `FilterFileSystem` in https://github.com/palantir/hadoop-crypto/pull/94 ?

ellisjoe (approved):

PR #102, opened 2018-01-09
Baseline 0.18.0

ellisjoe:


.baseline/copyright/999_palantir.txt
```diff
@@ -0,0 +1 @@
+(c) Copyright ${today.year} Palantir Technologies Inc. All rights reserved.
```
uschi2000: file shouldn't exist since it makes the palantir internal copyright the default one.

ellisjoe: removed

ellisjoe: do we have any other standard configurations for things like this other than `internal` and `external`? would be nice to flag this once since this file will come back every time we `./gradle updateBaselineConfig`

markelliot: I think we want things to default to the internal copyright.

ellisjoe: Agree that internal should be the default. Just want to be able to make a one time change so I don't screw up copyrights every time I run the `updateBaselineConfig` task

uschi2000:
Could do that in Baseline with code and config. This is not implemented today.

-------- Original Message --------

PR #104, opened 2018-01-17
Sr/fix distcp

sjrand:
Somewhere between 2.3.2 and 2.3.4, distcp from `hdfs://...` to `ehdfs://...` got broken in a few ways:

* The version of `create` that distcp uses doesn't know to write a key to the keystore -- need to have EFS call `keyStore.put`
* Need PathConvertingFS to convert the path for the version of `create` that it's using
* Need PathConvertingFS to convert the path for the version of `getFileChecksum` that it's using

This code works, but should probably understand what exactly in the 2.3.2 -> 2.3.4 bump caused these issues, thus the "don't merge" for now. Likely suspect seems to be the move to DelegatingFS, though it's not obvious to me why exactly. Will look more tomorrow, but won't complain if someone wants to figure it out while I'm asleep :)

Also it wasn't necessary to pass the `-skipcrccheck` flag to distcp on 2.3.2, but it is on this branch. Otherwise distcp gets confused that the checksums don't match (which makes sense given the unencrypted source files and encrypted destination files).

### DIFF
--- a/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
@@ -91,7 +91,7 @@ public final class EncryptedFileSystem extends DelegatingFileSystem {
         FSDataOutputStream encryptedStream =
                 fs.create(path, permission, overwrite, bufferSize, replication, blockSize, progress);
 
-        return storeKey(encryptedStream, path);
+        return encrypt(encryptedStream, path);
     }
 
     @Override
@@ -100,10 +100,10 @@ public final class EncryptedFileSystem extends DelegatingFileSystem {
         FSDataOutputStream encryptedStream =
                 fs.create(path, permission, flags, bufferSize, replication, blockSize, progress, checksumOpt);
 
-        return storeKey(encryptedStream, path);
+        return encrypt(encryptedStream, path);
     }
 
-    private FSDataOutputStream storeKey(FSDataOutputStream encryptedStream, Path path) throws IOException {
+    private FSDataOutputStream encrypt(FSDataOutputStream encryptedStream, Path path) throws IOException {
         KeyMaterial keyMaterial = SeekableCipherFactory.generateKeyMaterial(cipherAlgorithm);
         SeekableCipher cipher = SeekableCipherFactory.getCipher(cipherAlgorithm, keyMaterial);
 
--- a/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
@@ -27,11 +27,14 @@ import com.palantir.crypto2.keys.KeyStorageStrategy;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
+import java.util.EnumSet;
 import java.util.Optional;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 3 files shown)


sjrand:
Err ... for the issues with `create` at least, it looks like the breaks stem from having PathConvertingFS extend DelegatingFS instead of just FileSystem. So now distcp's version of `create` gets delegated all the way down to EFS, whereas previously it would just fall through to the version of `create` which PathConvertingFS overrides, and only get delegated down to EFS in that form. 

So sometimes not delegating FS calls breaks things, and sometimes delegating them breaks things.

Anyway, this fix seems reasonableish to me now that I think I know what's going on, so updating the title -- definitely open to alternatives though.

(Edit: And the checksum issues are probably caused by us actually checksumming now, instead of using the default impl which is `return null;`)

sjrand:
Tagging @ellisjoe / @uschi2000 for thoughts

ellisjoe:
seems fine. need tests for the methods added. It sounds like we should really be overriding every single `create` method as well?

sjrand:
Yeah, every other `create` method will have the same problem, though overriding all of them adds a lot of boilerplate code. I'm fine either way -- lemme know which way you want to go with it and will do.

ellisjoe:
Yeah this is fine. Don't need to do all of the create methods here. Filing a ticket for it though.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -88,6 +91,19 @@ public FSDataOutputStream create(Path path, FsPermission permission,
...
+    @Override
+    public FSDataOutputStream create(Path path, FsPermission permission, EnumSet<CreateFlag> flags, int bufferSize,
+            short replication, long blockSize, Progressable progress, ChecksumOpt checksumOpt) throws IOException {
+        FSDataOutputStream encryptedStream =
+                fs.create(path, permission, flags, bufferSize, replication, blockSize, progress, checksumOpt);
+
+        return storeKey(encryptedStream, path);
+    }
+
+    private FSDataOutputStream storeKey(FSDataOutputStream encryptedStream, Path path) throws IOException {
```
ellisjoe: `storeKey` -> `encrypt`

ellisjoe (commented):
should add tests for EncryptedFileSystem as well

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
sjrand: Done, though I think it's a bit confusing that now we're saying `return encrypt(encryptedStream, path);` in the create methods. I think the `encryptedStream` variable might not be named correctly since it's just the result of calling `fs.create`, i.e., the delegate won't actually do the encryption, right?

sjrand:
I wasn't sure how to test the EFS change since `EncryptedFileSystemTest` doesn't check each method the way `PathConvertingFileSystemTest` does -- looks like it only ever calls `create(Path)` rather than the ones it overrides. 

(Also left a reply under the `storeKey -> encrypt` comment which isn't visible w/o clicking "show outdated")

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: yeah should rename the variable as well. `outputStream` is fine.

ellisjoe:
merging. will fix the rename and add a test in another PR

ellisjoe (approved):

sjrand:
sweet, thanks

PR #106, opened 2018-01-24
Test newly overridden create method

ellisjoe:


hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -97,10 +97,10 @@ public FSDataOutputStream create(Path path, FsPermission permission,
     @Override
     public FSDataOutputStream create(Path path, FsPermission permission, EnumSet<CreateFlag> flags, int bufferSize,
             short replication, long blockSize, Progressable progress, ChecksumOpt checksumOpt) throws IOException {
-        FSDataOutputStream encryptedStream =
+        FSDataOutputStream outputStream =
```
sjrand, line 100: Same thing for the other `create` method above this one

sjrand (approved):
Just the one minor comment, otherwise lgtm

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: done

PR #109, opened 2018-02-14
Allow inheriting from EncryptedFileSystem

hsaraogi:
need it for allowing logical path to be passed in some cases.

@ltran @pnepywoda for SA.


pnepywoda:
can't you just delegate instead?

ltran:
i think you got the wrong person, might want to mention the person with similar handle as mine to get your review.

ash211:
@nartmal

hsaraogi:
I want a different KeyStorageStrategy and want to derive the path here too, so cant delegate.

markelliot:
your delegate can implement DelegatingFileSystem and inject an EncryptedFileSystem configured as you like, allowing you to override DelegatingFileSystem methods as desired. It's not practical for us to allow arbitrary override of the EncryptedFileSystem methods.

PR #110, opened 2018-02-21
Test unwrapping KeyMaterial with invalid key fails

ellisjoe:


crypto-keys/src/test/java/com/palantir/crypto2/keys/serialization/KeyMaterialsTest.java
```diff
@@ -73,6 +74,17 @@ public void testWrapAndUnwrap() {
         assertThat(unwrapped).isEqualTo(keyMaterial);
     }
 
+    @Test
+    public void testUnwrapFailsWhenUsedWithWrongKeyPair() {
+        KeyPair invalidKeyPair = TestKeyPairs.generateKeyPair();
+        byte[] wrapped = KeyMaterials.wrap(keyMaterial, keyPair.getPublic());
+
+        assertThatExceptionOfType(RuntimeException.class)
```
uschi2000, line 82: prefer this one over `assertThatThrownBy(()->foo()).isInstanceOf(RuntimexException.class)`?

uschi2000 (approved):

crypto-keys/src/test/java/com/palantir/crypto2/keys/serialization/KeyMaterialsTest.java
ellisjoe: I think this one reads a bit nicer. the other would maybe be nicer if it read like
`assertThatExceptionThrownBy(() -> foo()).isInstanceOf(RuntimeException.class)`? guess I just don't like `assertThatThrownBy`

Issue #93, opened 2017-09-22
Recursive delete support


SerialVelocity:
Sometime between 2.0.0-rc7 and 2.3.2 recursive deletes started throwing an exception about them being unsupported.

It would be nice if this was supported so that whole folders of encrypted data can be deleted easily. See internal backup solution PR 720.

jpfourny:
+1. We can't use this crypto library with Spark (and many other Hadoop based tools) without recursive delete support. Hadoop job cleanup uses recursive delete on temp directories.

ellisjoe:
@sandorw do you recall why we didn't want to support recursive deletes? the exception was added here: https://github.com/palantir/hadoop-crypto/pull/47/files

sandorw:
IIRC, the reason was that, while it's well supported for file based key storage, it's harder to implement for `InMemoryKeyStorageStrategy` without changing the current simple implementation. Since it wasn't clear we needed it at the time, we opted out of the extra complexity.

ellisjoe:
Fixed by https://github.com/palantir/hadoop-crypto/pull/111

ashrayjain:
@ellisjoe We are still getting this from `EncryptedFileSystem`. Specifically, this line:

https://github.com/palantir/hadoop-crypto/blob/develop/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java#L147

ellisjoe:
At the moment that is expected behavior. You should use the StandaloneEncryptedFileSystem if you're using the FileKeyStorageStrategy and need recursive deletes.

ashrayjain:
Cool, that's what we ended up doing. Thanks!

PR #111, opened 2018-02-27
Support deletes in StandaloneEncryptedFileSystem

ellisjoe:


ellisjoe:
Fixes https://github.com/palantir/hadoop-crypto/issues/93

sandorw (approved):

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -129,6 +131,14 @@ public boolean exists(Path path) throws IOException {
         return fs.exists(path);
     }
 
+    @Override
+    public boolean delete(Path path, boolean recursive) throws IOException {
+        // EncryptedFileSystem does not handle recursive deletes because it needs to rely on the KeyStorageStrategy.
+        // Here we know we are using the FileKeyStorageStrategy and can therefore delegate deletes to the FileSystem
+        // which will remove both the data and key material files.
+        return delegate.delete(path, recursive);
```
uschi2000, line 139: what if path is a file rather than a folder?

uschi2000: the key material won't get deleted, so this seems broken

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -73,15 +73,17 @@ public boolean apply(FileStatus status) {
      * Key mapping to the public/private key algorithm (ex: RSA).
      */
     public static final String KEY_ALGORITHM_CONF = "fs.efs.key.algorithm";
+
     private String encryptedScheme;
+    private FileSystem delegate;
```
uschi2000, line 78: should explain what this is and how it's different from `this.fs`.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
```diff
@@ -102,6 +103,49 @@ public void testReadWrite() throws IOException {
...
+        Path path1 = writeData(rootFolder);
+        Path path2 = writeData(rootFolder);
+
+        assertThat(efs.exists(path1)).isTrue();
+        assertThat(efs.exists(path2)).isTrue();
+
+        efs.delete(new Path(rootFolder.getAbsolutePath()), true);
+
+        assertThat(efs.exists(path1)).isFalse();
+        assertThat(efs.exists(path2)).isFalse();
```
uschi2000, line 135: assertion about rootFolder missing

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
```diff
@@ -102,6 +103,49 @@ public void testReadWrite() throws IOException {
...
+    }
+
+    @Test
+    public void testRecursiveDelete() throws IOException {
+        File rootFolder = folder.newFolder();
+
+        Path path1 = writeData(rootFolder);
+        Path path2 = writeData(rootFolder);
+
+        assertThat(efs.exists(path1)).isTrue();
```
uschi2000, line 129: move these to writeData?

uschi2000: nah, actually fine here.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
```diff
@@ -102,6 +103,49 @@ public void testReadWrite() throws IOException {
...
+        assertThat(efs.exists(path1)).isFalse();
+        assertThat(efs.exists(path2)).isFalse();
+    }
+
+    private Path writeData(File rootFolder) throws IOException {
+        // Write encrypted data
+        java.nio.file.Path filePath = Files.createTempFile(rootFolder.toPath(), "prefix", "suffix");
+        Path newPath = new Path(filePath.toAbsolutePath().toString());
+        OutputStream os = efs.create(path);
+        IOUtils.write("data", os);
```
uschi2000, line 143: think you meant to write `big data` here.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
```diff
@@ -102,6 +103,49 @@ public void testReadWrite() throws IOException {
         assertThat(rawFs.exists(new Path(path + FileKeyStorageStrategy.EXTENSION))).isTrue();
     }
 
+    @Test
+    public void testDelete() throws IOException {
+        File rootFolder = folder.newFolder();
+
+        Path path1 = writeData(rootFolder);
+        Path path2 = writeData(rootFolder);
```
uschi2000, line 111: this seems a bit bogus: you're verifying that `delete` doesn't touch a totally unrelated path?

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
```diff
@@ -102,6 +103,49 @@ public void testReadWrite() throws IOException {
...
+
+        assertThat(efs.exists(path1)).isTrue();
+        assertThat(efs.exists(path2)).isTrue();
+
+        efs.delete(new Path(path1.toString()), false);
+
+        assertThat(efs.exists(path1)).isFalse();
```
uschi2000, line 118: - should verify that you write to the same location again after deleting from it, I think.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
ellisjoe: removed in the follow up PR

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
ellisjoe: fixed in the follow up

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
ellisjoe: added in follow up

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
ellisjoe: done in follow up

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
ellisjoe: don't think that's relevant here? in my mind we've shown that `delete` puts the FileSystem back into a state where a file and it's key material at a given path do not exist. Our `create` tests should verify that we can create a new encrypted file from that state.

uschi2000: ok, now that you're actually verifying that, that's fine :)

Issue #100, opened 2017-12-31
Support renaming directories


jpfourny:
For this to work with tools like Spark, support is required for renaming directories. Currently, EncryptedFileSystem naively assumes the path being renamed is a regular file and barfs when it cannot find a key material for the directory. This is related to "Recursive delete support #93", in that there appears to be a general lack of consideration for recursive directory operations.

ellisjoe:
The issue with the batch operations is that we don't have a way to atomically update every `KeyStorageStrategy`. If you're just using the `FileKeyStorageStrategy` we could potentially make a specific `EncryptedFileSystem` that uses the `FileKeyStorageStrategy` directly rather than being pluggable. It can then rely on the `FileSystem` rename and delete calls rather than the `KeyStorageStrategy` methods to move `KeyMaterial` around.

ellisjoe:
@jpfourny released https://github.com/palantir/hadoop-crypto/releases/tag/2.4.0 which should fix your issues

PR #112, opened 2018-02-27
Support rename in StandaloneEncryptedFileSystem

ellisjoe:


ellisjoe:
fixes https://github.com/palantir/hadoop-crypto/issues/100

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -139,6 +139,13 @@ public boolean delete(Path path, boolean recursive) throws IOException {
         return delegate.delete(path, recursive);
     }
 
+    @Override
+    public boolean rename(Path src, Path dst) throws IOException {
+        // Similarly to delete, rename is able to rely on the underlying FileSystem properly moving data and key
+        // material files.
+        return delegate.rename(src, dst);
```
uschi2000: this is broken when `src` is a file, not a folder

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -139,6 +139,18 @@ public boolean delete(Path path, boolean recursive) throws IOException {
...
 
+    @Override
+    public boolean rename(Path src, Path dst) throws IOException {
+        // The EncryptedFileSystem only works with files to allow the KeyStorageStrategy interface to be simple.
+        // Because we use the FileKeyStorageStrategy we know that we can also support operations on directories
+        // and therefore delegate to the underlying FileSystem in that case.
+        if (fs.isFile(src)) {
+            return fs.rename(src, dst);
+        } else {
+            return delegate.rename(src, dst);
```
ellisjoe: @uschi2000 how do you feel about this logic and the comment? think it's clear enough and can replace what I've got in the `delete` implementation.

uschi2000: I would make the leap from "operations on directories" to "can delegate" more explicit: Since StandaloneEncryptedFileSystem uses a FileKeyStorageStrategy, the delegate delete/rename call on folders deletes both the payload files and the adjacent encryption materials. For files ...

remember, you're not writing the docs for yourself, but for someone else who potential has no clue how this things works.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -139,6 +139,18 @@ public boolean delete(Path path, boolean recursive) throws IOException {
         return delegate.delete(path, recursive);
     }
 
+    @Override
+    public boolean rename(Path src, Path dst) throws IOException {
+        // Since StandaloneEncryptedFileSystem uses a FileKeyStorageStrategy, the delegate rename call on folders
+        // renames both the payload files and the adjacent encryption materials. For files we can continue to allow
```
uschi2000: -`continue to allow`

uschi2000 (approved):

PR #113, opened 2018-02-27
Delete key material on non recursive delete

ellisjoe:


hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -135,7 +137,11 @@ public boolean exists(Path path) throws IOException {
     public boolean delete(Path path, boolean recursive) throws IOException {
         // EncryptedFileSystem does not handle recursive deletes because it needs to rely on the KeyStorageStrategy.
```
uschi2000: took me a minute to understand this comment without prior context, so maybe worth explaining a bit more thoroughly: `In contrast to EncryptedFileSystem (which works with arbitrary KeyStorageStrategies), a StandaloneEncryptedFileSystem always uses a FileKeyStorageStrategy which stores key materials as files adjacent to payload files. This allows us to implement recursive deletion of folders (and rename, see below) by simply deleting (or renaming) the folder in the underlying delegate file system: it deletes (or renames) both the payload files and the adjacent key material files. When deleting files, we need to explicitly delete the payload file and the key material file.`

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -135,7 +137,11 @@ public boolean exists(Path path) throws IOException {
...
         // Here we know we are using the FileKeyStorageStrategy and can therefore delegate deletes to the FileSystem
-        // which will remove both the data and key material files.
+        // which will remove both the data and key material files. If the path is a directory then the recursive delete
+        // will handle removing the key material files, otherwise we will explicitly delete the key material here.
+        if (fs.isFile(path)) {
+            keyStore.remove(path.toString());
```
uschi2000: does the order of the two delete calls matter? in theory, if you fail the `delegate.delete` call then `exists()` returns true for the file, but it still can't be read. maybe doesn't matter.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
```diff
@@ -100,46 +100,60 @@ public void testReadWrite() throws IOException {
...
         assertThat(efs.exists(path2)).isFalse();
+        assertThat(rawFs.exists(path2)).isFalse();
+        assertThat(rawFs.exists(keyMaterialPath(path2))).isFalse();
     }
 
     private Path writeData(File rootFolder) throws IOException {
         // Write encrypted data
         java.nio.file.Path filePath = Files.createTempFile(rootFolder.toPath(), "prefix", "suffix");
         Path newPath = new Path(filePath.toAbsolutePath().toString());
-        OutputStream os = efs.create(path);
```
uschi2000, line 142: oh, this is majestic

uschi2000: was wondering why the non-recursive delete worked before ...

uschi2000 (approved):
some nits, otherwise +1

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystemTest.java
ellisjoe: well the file actually does get created by `Files.createTempFile` above which made this even more insidious

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
ellisjoe: don't think it matters either way? think this ends up reading a bit more nicely though since I don't ned to keep around the result of the `delete.delete` call

uschi2000: works for me

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
ellisjoe: Gonna think about this...seems like something is a bit off if we need so much explanation for a single method override but maybe this is ok.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -133,10 +135,14 @@ public boolean exists(Path path) throws IOException {
 
     @Override
     public boolean delete(Path path, boolean recursive) throws IOException {
-        // EncryptedFileSystem does not handle recursive deletes because it needs to rely on the KeyStorageStrategy.
-        // Here we know we are using the FileKeyStorageStrategy and can therefore delegate deletes to the FileSystem
-        // which will remove both the data and key material files.
-        return delegate.delete(path, recursive);
+        // Since StandaloneEncryptedFileSystem uses a FileKeyStorageStrategy, the delegate delete call on folders
+        // deletes both the payload files and the adjacent encryption materials. For files we can continue to allow
```
uschi2000: -allow

uschi2000: actually -`continue to allow`

uschi2000 (approved):

PR #114, opened 2018-02-28
Prevent docker build from writing build files as root

ellisjoe:


ellisjoe:
Builds started failing at the bintray publish phase due to incorrect perms on files. This appears to be due to mounting the working directory into docker, which then creates files as root and renders them unreadable by future build stages. This copies the working directory into the docker container to prevent the docker and non-docker build steps from touching the same files.

j-baker (approved):

PR #115, opened 2018-02-28
Older versions of nebula to fix publishing

ellisjoe:


ellisjoe:
@gracew @iamdanfox the newer versions of nebula broke publishing. jars weren't published and the pom was missing nearly all of the entries. any ideas what the issue is here?

ellisjoe:
merging to get things publishing

gracew:
not sure -- I've seen this in the past when using gradle 3.x + the newer nebula plugin versions, but your version of 4.2.1 looks fine.

schlosna:
@ellisjoe You probably have to `apply plugin: "nebula.maven-resolved-dependencies"` per https://github.com/nebula-plugins/nebula-dependency-recommender-plugin#9-notes-on-poms-generated-by-gradle-maven-publish

> Notes on POMs Generated by Gradle maven-publish
> Gradle requires that version numbers are present in the dependencies block to create a valid POM file that includes version numbers. To fix the issue this causes when using the dependency-recommender plug-in, apply the nebula.maven-resolved-dependencies plug-in from the nebula-publishing-plugin set.

I had to do this for https://github.com/palantir/tritium/commit/9a479e60dd3b36b7009af060b6c4f7c5cf9634d6

ellisjoe:
@schlosna yeah I was applying that in `gradle/publish.gradle`. Turned out to be that I needed to add a `publishing` block, which interestingly fails on older versions of nebula. Real fix: https://github.com/palantir/hadoop-crypto/pull/116

PR #116, opened 2018-03-01
Fix pom publishing

ellisjoe:


ellisjoe:
@gracew is there any way we can prevent the silent failures? in older versions of nebula this was the default configuration, so upgrading just meant poms were essentially empty.

Issue #117, opened 2018-04-19
EncryptedFileSystem does not support recursive delete


TheBeruriahIncident:
https://github.com/palantir/hadoop-crypto/blob/develop/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java#L147

I may be missing some complexity, but it seems like recursive delete would just be a few more lines. Is there something I'm missing that would make this more complicated? Would a PR to support this be accepted?

TheBeruriahIncident:
Just realized (it's not in the README but) there's a StandaloneEncryptedFileSystem that already supports recursive delete. I don't really see why to use one or the other, other than that StandaloneEncryptedFileSystem has a kind of weird requirement to insert an `e` at the start of the uris that it then removes.

ellisjoe:
We need to make `KeyStorageStrategy` expose recursive delete and rename options in order for EncryptedFileSystem to support recursive delete and move. This is unlikely to happen any time soon, so the recommendation in the short term is either use StandaloneEncryptedFileSystem or use the EncryptedFileSystem with the FileKeyStorageStrategy and recursively delete or move on the underlying filesystem.

rahij:
@ellisjoe is it still true that exposing recursive delete and rename options in `KeyStorageStrategy` won't happen anytime soon?

ellisjoe:
@rahij we should talk about your need for recursive delete. Hasn't been a priority thus far.

rahij:
Synced offline with Joe, we have a workaround for the internal library.

PR #118, opened 2018-04-19
Add support for recursive delete in EncryptedFileSystem

TheBeruriahIncident:
Fixes https://github.com/palantir/hadoop-crypto/issues/117

We want to be able to delete folders of many items. CC @eraboin

If this change seems generally reasonable, I can add more tests

### DIFF
--- a/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
@@ -27,12 +27,14 @@ import com.palantir.crypto2.keys.KeyStorageStrategy;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
+import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.Optional;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.CreateFlag;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Options.ChecksumOpt;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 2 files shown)


hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -143,8 +144,16 @@ private void tryRemoveKey(Path path) {
...
-        if (recursive) {
-            throw new UnsupportedOperationException("EncryptedFileSystem does not support recursive deletes");
+        if (recursive && isDirectory(path)) {
+            FileStatus[] children = listStatus(path);
+
+            for (FileStatus child : children) {
+                boolean deleteSuccessful = delete(child.getPath(), true);
+
+                if (!deleteSuccessful) {
+                    return false;
```
eraboin, line 154: The definition of `delete` in the [Hadoop documentation](http://hadoop.apache.org/docs/r2.7.5/api/org/apache/hadoop/fs/FilterFileSystem.html#delete-org.apache.hadoop.fs.Path-boolean-) does not specify whether we should stop immediately here, or continue with the rest of the delete. If we're modeling this after `rm -rf` then I'd expect it to continue deleting what it can though, right?

Not sure if there's a reference impl elsewhere we can look at.

TheBeruriahIncident: Easy to do either, but yeah, I'm unsure what behavior is preferable. RawLocalFileSystem continues and eventually returns false if anything failed. FTPFileSystem continues and ignores any intermediate issues (maybe a bug). DistributedFileSystem is hard to follow, but seems to atomically destroy everything, making this is a non-issue. S3FileSystem matches the behavior of this PR, returning early if anything fails.

ellisjoe: I'd really prefer to avoid supporting recursive deletes within EncryptedFileSystem until we can do so efficiently. I'd propose that you do:
```
RemoteIterator<LocatedFileStatus> files = efs.listFiles(path, true);
while (files.hasNext()) {
    efs.delete(files.next().getPath(), false);
}
```
as a workaround for the time being.

PR #123, opened 2018-05-08
Offer serializer that uses symmetric key encryption for wrapping oper…

ellisjoe:
…ations

### DIFF
--- a/crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
+++ b/crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
@@ -1,0 +1,140 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.crypto2.keys.serialization;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
+import com.palantir.crypto2.keys.KeyMaterial;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.IvParameterSpec;
+
+/**
+ * Serializer for wrapping and unwrapping {@link KeyMaterial}. The {@link #wrap} method returns the KeyMaterial
+ * serialized as follows, which is the same format the {@link #unwrap} method expects:
+ *
+ * <pre>
+ *  +--------------------------------------------------------------+
+ *  | version | wrapping iv | key algorithm length | key algorithm |
+ *  |   byte  |    byte[]   |          int         |     byte[]    |
+ *  +--------------------------------------------------------------+
+ *  +-------------------------------------------------------+
+ *  | wrapped key length | wrapped key | iv length |   iv   |
+ *  |        int         |    byte[]   |    int    | byte[] |
+ *  +-------------------------------------------------------+
+ * </pre>
+ */
+enum SymmetricKeySerializerV3 implements SymmetricKeySerializer {
+    INSTANCE;
+
+    private static final String AES_CTR_NO_PADDING = "AES/CTR/NoPadding";
+    private static final int IV_SIZE = 16;
+
+    static final int VERSION = 3;
+
+    @Override
+    public byte[] wrap(KeyMaterial keyMaterial, SecretKey key) {
+        ByteArrayOutputStream byteStream = new ByteArrayOutputStream();
+        DataOutputStream stream = new DataOutputStream(byteStream);
+
+        byte[] wrappingIv = KeyMaterials.generateIv(IV_SIZE);
+        Cipher keyWrappingCipher = getCipher(Cipher.WRAP_MODE, key, wrappingIv);
+        SecretKey secretKey = keyMaterial.getSecretKey();
+
+        try {
+            stream.write(VERSION);
+
+            stream.write(wrappingIv);
+
+            String keyAlgorithm = secretKey.getAlgorithm();
+            stream.writeInt(keyAlgorithm.length());
+            stream.write(keyAlgorithm.getBytes(StandardCharsets.UTF_8));
+
+            byte[] encryptedKey = keyWrappingCipher.wrap(secretKey);
+            stream.writeInt(encryptedKey.length);
+            stream.write(encryptedKey);
+
+            byte[] iv = keyMaterial.getIv();
+            stream.writeInt(iv.length);
+            stream.write(iv);
+
+            stream.close();
+            return byteStream.toByteArray();
+        } catch (IOException | InvalidKeyException | IllegalBlockSizeException e) {
+            throw Throwables.propagate(e);
+        }
+    }
+
+    @Override
+    public KeyMaterial unwrap(byte[] wrappedKeyMaterial, SecretKey key) {
+        DataInputStream stream = new DataInputStream(new ByteArrayInputStream(wrappedKeyMaterial));
+
+        try {
+            int version = stream.read();
+            Preconditions.checkArgument(VERSION == version,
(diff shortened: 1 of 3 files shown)


uschi2000:
@cbarcenas

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
```diff
@@ -0,0 +1,140 @@
...
+ *  +-------------------------------------------------------+
+ * </pre>
+ */
+enum SymmetricKeySerializerV3 implements SymmetricKeySerializer {
+    INSTANCE;
+
+    private static final String AES_CTR_NO_PADDING = "AES/CTR/NoPadding";
+    private static final int IV_SIZE = 16;
+
+    static final int VERSION = 3;
```
uschi2000, line 58: private?

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
```diff
@@ -0,0 +1,140 @@
...
+            stream.write(encryptedKey);
+
+            byte[] iv = keyMaterial.getIv();
+            stream.writeInt(iv.length);
+            stream.write(iv);
+
+            stream.close();
+            return byteStream.toByteArray();
+        } catch (IOException | InvalidKeyException | IllegalBlockSizeException e) {
+            throw Throwables.propagate(e);
```
uschi2000: throw runtimeexception with message

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
```diff
@@ -0,0 +1,140 @@
...
+
+            int ivLength = stream.readInt();
+            byte[] iv = new byte[ivLength];
+            stream.readFully(iv);
+
+            String algorithm = new String(algorithmBytes, StandardCharsets.UTF_8);
+            SecretKey secretKey = (SecretKey) keyUnwrappingCipher.unwrap(secretKeyBytes, algorithm, Cipher.SECRET_KEY);
+            return KeyMaterial.of(secretKey, iv);
+        } catch (InvalidKeyException | NoSuchAlgorithmException | IOException e) {
+            throw Throwables.propagate(e);
```
uschi2000: ditto

crypto-keys/src/test/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3Test.java
```diff
@@ -0,0 +1,40 @@
...
+
+public final class SymmetricKeySerializerV3Test {
+
+    private static final SymmetricKeySerializerV3 SERIALIZER = SymmetricKeySerializerV3.INSTANCE;
+    private static final String KEY_ALG = "AES";
+    private static final int KEY_SIZE = 128;
+    private static final int IV_SIZE = 16;
+
+    @Test
+    public void test() {
```
uschi2000: should test:
- uses fresh IV (i.e., `encrypt(foo).notEquals(encrypt(foo))`

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
ellisjoe: going to expose this as part of `KeySerializers` which does the multiplexing base one version for you

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
ellisjoe: done

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
ellisjoe: done

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
uschi2000: but there's a `getVersion()` method?

crypto-keys/src/test/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3Test.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Test
+    public void testNewIvUsed() {
+        KeyMaterial wrappingKeyMaterial = KeyMaterials.generateKeyMaterial(KEY_ALG, KEY_SIZE, IV_SIZE);
+        KeyMaterial keyMaterial = KeyMaterials.generateKeyMaterial(KEY_ALG, KEY_SIZE, IV_SIZE);
+
+        byte[] wrapped1 = SERIALIZER.wrap(keyMaterial, wrappingKeyMaterial.getSecretKey());
+        byte[] wrapped2 = SERIALIZER.wrap(keyMaterial, wrappingKeyMaterial.getSecretKey());
+        assertThat(wrapped1).isNotEqualTo(wrapped2);
```
uschi2000, line 48: did you double-check that this compares the data, not the pointers?

uschi2000 (approved):

crypto-keys/src/test/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3Test.java
ellisjoe: the pointers check would only pass if we were passing back a static byte[]. it does check the contents of the arrays here, but only because they are not `equals`

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
```diff
@@ -0,0 +1,143 @@
...
+    public KeyMaterial unwrap(byte[] wrappedKeyMaterial, SecretKey key) {
+        DataInputStream stream = new DataInputStream(new ByteArrayInputStream(wrappedKeyMaterial));
+
+        try {
+            int version = stream.read();
+            Preconditions.checkArgument(VERSION == version,
+                    "Invalid serialization format version. Expected %s but found %s", VERSION, version);
+
+            byte[] wrappingIv = new byte[IV_SIZE];
+            stream.read(wrappingIv);
```
schlosna, line 103: This needs to be `readFully` as `read` may not read all bytes

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
```diff
@@ -0,0 +1,143 @@
...
+        DataOutputStream stream = new DataOutputStream(byteStream);
+
+        byte[] wrappingIv = KeyMaterials.generateIv(IV_SIZE);
+        Cipher keyWrappingCipher = getCipher(Cipher.WRAP_MODE, key, wrappingIv);
+        SecretKey secretKey = keyMaterial.getSecretKey();
+
+        try {
+            stream.write(VERSION);
+
+            stream.write(wrappingIv);
```
schlosna, line 72: realize its constant, but should we write the IV length before its bytes?

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
ellisjoe: thanks. @uschi2000 should probably have an error-prone check for this.

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
ellisjoe: didn't include that in this case given this IV is controlled by this wrapping function and not provided by the caller (which is the case with the other IV). think if we want to change the IV length or the algorithm, it's nicer to just bump the version we're writing rather than making this format more generic.

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
uschi2000: What would you want to check? That you have to use the return value of read?

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
uschi2000: +1, think it's fine to bump the version in that case.

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
schlosna: Yep, IntelliJ's static analysis flags this, see also https://wiki.sei.cmu.edu/confluence/display/java/FIO10-J.+Ensure+the+array+is+filled+when+using+read%28%29+to+fill+an+array

This specific case would likely never hit an issue since we're reading from ByteArrayInputStream, but if someone copy/pasted to read from network or disk, sad times may ensue.

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
schlosna: fine by me

PR #125, opened 2018-05-09
read -> readyFully

ellisjoe:


schlosna (approved):
LGTM, thanks @ellisjoe

This specific case is safe since we’re doing read from ByteArrayInputStream so it will always do the right thing, but agree we should have error prone check

ellisjoe:
yeah glad you caught this even though it would've worked in practice. definitely one of those easy to miss but sad to debug/fix things.

PR #127, opened 2018-05-09
Use AES/GCM for symmetric key wrapping

ellisjoe:


ellisjoe:
Ran some benchmarks here: https://gist.github.palantir.build/jellis/aa80c35a565dc5b1e50de2d2f731d7c9

and found that perf wasn't a real issue with values as small as the keys we're wrapping. Think the small perf hit is worth the authentication benefits we'll get

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
```diff
@@ -52,9 +52,11 @@
 enum SymmetricKeySerializerV3 implements SymmetricKeySerializer {
     INSTANCE;
 
-    private static final String AES_CTR_NO_PADDING = "AES/CTR/NoPadding";
-    private static final int IV_SIZE = 16;
-
+    private static final String AES_CTR_NO_PADDING = "AES/GCM/NoPadding";
```
markelliot: `AES_CTR_NO_PADDING` -> `AES_GCM_NO_PADDING`

ellisjoe: done. thanks

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
```diff
@@ -52,9 +52,11 @@
 enum SymmetricKeySerializerV3 implements SymmetricKeySerializer {
     INSTANCE;
 
-    private static final String AES_CTR_NO_PADDING = "AES/CTR/NoPadding";
-    private static final int IV_SIZE = 16;
-
+    private static final String AES_GCM_NO_PADDING = "AES/GCM/NoPadding";
+    private static final int IV_SIZE = 12;
+    // 128 bit tag length as recommended by:
+    // https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf
```
uschi2000, line 58: @cbarcenas could you take a look at this, please?

uschi2000 (approved):
looks good to me, but would suggest to have @cbarcenas take a peek

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
```diff
@@ -52,9 +52,11 @@
 enum SymmetricKeySerializerV3 implements SymmetricKeySerializer {
     INSTANCE;
 
-    private static final String AES_CTR_NO_PADDING = "AES/CTR/NoPadding";
-    private static final int IV_SIZE = 16;
-
+    private static final String AES_GCM_NO_PADDING = "AES/GCM/NoPadding";
+    private static final int IV_SIZE = 12;
```
schlosna, line 56: Realize no-one is probably using this yet, but we're breaking compat via algorithm & IV size switch. should we embed in serialized format?

ellisjoe: I'd personally prefer to rely on the version byte determining the algorithm and iv size. this isn't released yet which is the only reason I didn't bump the version.

ellisjoe:
merging but will wait for @cbarcenas and a confirmation on my response to @schlosna before releasing

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/SymmetricKeySerializerV3.java
cbarcenas: This is standard - IVs in GCM mode are usually 92 bits wide. Inside the GCM construction, a 32-bit counter suffix is appended to the input IV to form a 128-bit-wide input to the AES block function.

cbarcenas: It's probably worth updating the inline comment to better clarify this

ellisjoe: this comment is about the authentication tag, not the IV. I can add your comment to the IV_SIZE constant if you'd like.

PR #128, opened 2018-05-15
Decryption tests use StreamFactory

ellisjoe:


ellisjoe:
`CryptoStreamFactory` is really the api we want consumers to use so switch over the tests to use that directly. this should make sense given all of the stream construction code is removed since it is duplicated from `CryptoStreamFactory`

uschi2000 (approved):
sure

PR #160, opened 2018-11-27
Wrap with CBC for symmetric, not GCM

j-baker:
GCM is inferior to CBC here for two reasons.

1. GCM is much more wholly broken in the case that IV reuse happens for
a key.
2. GCM has a smaller IV (96 bits vs 128 bits) than CBC.

Or, to put it in terms of NIST's recommendations, you shouldn't use GCM
for more than 2^32 encryptions with a single key,
https://csrc.nist.gov/publications/detail/sp/800-38d/final (section 8.3).

Which the current setup encourage.

By using CBC instead, the birthday paradox becomes a lot less likely to
be hit (16 more bits of entropy in the IV), and the cost of hitting it
is much less drastic (you can only work out whether two keys have a
common prefix, which isn't useful)

### DIFF
--- a/crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/CipherSymmetricKeySerializer.java
+++ b/crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/CipherSymmetricKeySerializer.java
@@ -1,0 +1,133 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.crypto2.keys.serialization;
+
+import com.google.common.base.Preconditions;
+import com.palantir.crypto2.keys.KeyMaterial;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.SecretKey;
+
+/**
+ * Serializer for wrapping and unwrapping {@link KeyMaterial}. The {@link #wrap} method returns the KeyMaterial
+ * serialized as follows, which is the same format the {@link #unwrap} method expects:
+ *
+ * <pre>
+ *  +--------------------------------------------------------------+
+ *  | version | wrapping iv | key algorithm length | key algorithm |
+ *  |   byte  |    byte[]   |          int         |     byte[]    |
+ *  +--------------------------------------------------------------+
+ *  +-------------------------------------------------------+
+ *  | wrapped key length | wrapped key | iv length |   iv   |
+ *  |        int         |    byte[]   |    int    | byte[] |
+ *  +-------------------------------------------------------+
+ * </pre>
+ */
+public final class CipherSymmetricKeySerializer implements SymmetricKeySerializer {
+    private final int ivSize;
+    private final int version;
+    private final CipherFactory cipherFactory;
+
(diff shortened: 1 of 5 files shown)


j-baker:
Also, the risk is definitely low here; it would rely on an attacker's key being one of the two in the birthday paradox (given that a non-attacker would only be encrypting random bytes); but in a system where you can store ~2^48 things an attacker can probably store ~2^48 things.

ellisjoe:
My concern here is that CBC won't do the same authentication that GCM does. Looking around a bit, it also appears that GCM is what others (Amazon) are using for key wrapping. I'm looking a bit more into the specifics of how AWS KMS does this, but from their whitepape (https://d1.awsstatic.com/whitepapers/KMS-Cryptographic-Details.pdf):
```
Key Derivation Functions
A key derivation function is used to derive additional keys from an initial secret
or key. AWS KMS uses a key derivation function (KDF) to derive per-call keys for
every encryption under a CMK. All KDF operations use the KDF in counter mode
[7] using HMAC [FIPS197][8] with SHA256 [FIPS180] [9]. The 256-bit derived
key is used with AES-GCM to encrypt or decrypt customer data and keys.
```

ellisjoe:
The relevant NIST KDF documentation which includes a description of the mentioned KDF in counter mode: https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-108.pdf

And the AWS Encryption SDK doc just so I remember where it's at: https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/encryption-sdk-developer-guide.pdf

j-baker:
So I'm in agreement that something like the AWS thing makes sense. I had personally looked at something like AES-GCM-SIV as providing a similar primitive (though it's only in draft). My goal here is that we do effectively zero definition of our own crypto primitives; I don't think this repo should contain any home-derived crypto.

Effectively; I'd be happy to pull in the AWS Encryption SDK and use it, but I don't think we want to pull in BouncyCastle. At the same time, I'd also be happy to pull in https://github.com/codahale/aes-gcm-siv (which has no crypto), but that's only in draft.

ellisjoe:
So the AWS SDK doesn't do the KDF bit of the key generation, KMS does because it needs to store some state about the counter, so that won't help.

I don't think you want aes-gcm-siv, particularly because in the readme the first thing it say is that it's not ready.

CBC + HMAC should be ok, but is still not used in general:
https://blog.cloudflare.com/padding-oracles-and-the-decline-of-cbc-mode-ciphersuites/
https://docs.microsoft.com/en-us/dotnet/standard/security/vulnerabilities-cbc-mode

j-baker:
Right, that's why I didn't use AES-GCM-SIV in this PR. Could also use AES-SIV, but that has no trustworthy implementation in Java (by trustworthy I mean 'in javax.crypto').

Basically: Don't think there are any relevant issues with CBC in this use case (no padding oracle attacks possible), whereas GCM does have one (we're not using it in a standards compliant way). Ironically given what we're encrypting, block cipher usage doesn't matter, provided the key goes through the encryption function. While I agree that authentication is in general advantageous, I don't think it's critical for this use case (where in particular this typically is in lieu of RSA encryption which is _not_ authenticated in any way).

As yet another alternative, would you accept me using GCM, but with a key derived through PBKDF2 with 16 random bytes used as the salt? We'd have to store an extra IV type thing. Feels more like rolling our own crypto.

Basically:

I challenge that authentication is necessary here, though it is a nice to have. That said, we have a few ways of combining authentication with this in fairly sensible ways.

ellisjoe:
Alright I'm on board with this given we're not doing authentication anywhere and mention that in the readme. Think it's worth coming back here and adding authentication where possible.

ellisjoe (approved):

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/CipherSymmetricKeySerializer.java
```diff
@@ -0,0 +1,133 @@
...
+ *  | version | wrapping iv | key algorithm length | key algorithm |
+ *  |   byte  |    byte[]   |          int         |     byte[]    |
+ *  +--------------------------------------------------------------+
+ *  +-------------------------------------------------------+
+ *  | wrapped key length | wrapped key | iv length |   iv   |
+ *  |        int         |    byte[]   |    int    | byte[] |
+ *  +-------------------------------------------------------+
+ * </pre>
+ */
+public final class CipherSymmetricKeySerializer implements SymmetricKeySerializer {
```
ellisjoe, line 48: can we make this package private

crypto-keys/src/main/java/com/palantir/crypto2/keys/serialization/CipherSymmetricKeySerializer.java
```diff
@@ -0,0 +1,133 @@
...
+    private final int ivSize;
+    private final int version;
+    private final CipherFactory cipherFactory;
+
+    public CipherSymmetricKeySerializer(int ivSize, int version, CipherFactory cipherFactory) {
```
ellisjoe, line 53: same

j-baker:
Thanks!

PR #272, opened 2019-12-05
log error

yifeih:


markelliot (approved):

ellisjoe:
pretty sure this stack trace will always be exactly the same and end up just filling up logs since it'll get logged every time you open a file.

what info are you hoping to get out of the stacktrace?

yifeih:
Ah i see, i wanted it to diagnose some issues we were seeing with our s3 uploads, but i don't think we need this anymore. Closing.

ellisjoe:
Cool, let me know if you run into any other issues and I can try to get the stacktrace printed out.

PR #278, opened 2020-01-11
Log the exception when failing to create a ApacheCtrDecryptingSeekableInput

mattsills:
Log the exception when failing to create a `ApacheCtrDecryptingSeekableInput`


ellisjoe:
Had a similar request here: https://github.com/palantir/hadoop-crypto/pull/272

As mentioned there the exception is going to be the same every time so not sure how valuable it's going to be to log it. If we do want to log this, we should likely do it at `trace` rather than `info` to avoid blowing up logs.

mattsills:
What exception will it always be?

PR #281, opened 2020-01-13
Improvement: Use commons-crypto published from palantir fork

robert3005:


ellisjoe (approved):
Will release an RC and test internally before shipping more broadly given the amount of time between releases here: https://github.com/apache/commons-crypto/releases

PR #302, opened 2020-01-28
log exception when CryptoStreamFactory is unable to use fast codepath

iamdanfox:
## Before this PR

Looking in our internal log searching app for `origin:*CryptoStreamFactory*`, I found a _ton_ of services are not hitting the fast codepath. (Including our data ingestion coordinator on our dogfooding deployment).

> Unable to initialize cipher with OpenSSL, falling back to JCE implementation

This is somewhat expected for spark modules, but weird for regular cloud deployed services. @robert3005's hypothesis is just that OpenSSL is symlinked to the wrong place on those boxes, but it's hard to be certain without seeing the exception.

## After this PR
==COMMIT_MSG==
When CryptoStreamFactory is unable to use apache commons-crypto's fast native codepath, the WARN log line now includes the exception
==COMMIT_MSG==

## Possible downsides?
- more noise in logs?

### DIFF
--- a/crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
+++ b/crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
@@ -105,7 +105,8 @@ public final class CryptoStreamFactory {
     private static void maybeLogException(IOException exception) {
         long now = System.currentTimeMillis();
         long prev = mostRecentLoggedExceptionTimestamp.get();
-        if (now - prev > DELAY_BETWEEN_LOGGED_EXCEPTIONS && mostRecentLoggedExceptionTimestamp.compareAndSet(prev, now)) {
+        if (now - prev > DELAY_BETWEEN_LOGGED_EXCEPTIONS
+                && mostRecentLoggedExceptionTimestamp.compareAndSet(prev, now)) {
             log.warn("Exception for stack trace", exception);
         }
     }
--- a/crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
+++ b/crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
@@ -60,7 +60,7 @@ public final class CryptoStreamFactory {
         try {
             return new ApacheCtrDecryptingSeekableInput(encryptedInput, keyMaterial);
         } catch (IOException e) {
-            log.warn("Unable to initialize cipher with OpenSSL falling back to JCE implementation");
+            log.warn("Unable to initialize cipher with OpenSSL falling back to JCE implementation", e);
             return new DecryptingSeekableInput(encryptedInput, SeekableCipherFactory.getCipher(algorithm, keyMaterial));
         }
     }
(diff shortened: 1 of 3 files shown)


robert3005:
👍

ellisjoe:
alright, given this is the third time this has been requested let's go ahead and log this. @iamdanfox can we add a second log line here at debug level that includes the exception? I'd prefer not to log this at warn level since it'll just pollute logs aggressively

Other requests: https://github.com/palantir/hadoop-crypto/pull/272 and https://github.com/palantir/hadoop-crypto/pull/278

iamdanfox:
But part of the reason I think we've had this regress recently (and drop a bunch of performance) is because people don't notice these log lines.

As an alternative, how about just logging the exception at most once per minute?

ellisjoe:
Alright, one other proposal, given I don't think the exception is actually going to be helpful: what if we include instructions on how to deal with the error in the readme?

crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
```diff
@@ -91,11 +94,26 @@ static OutputStream encrypt(OutputStream output, KeyMaterial keyMaterial, String
...
-            log.warn("Unable to initialize cipher with OpenSSL, falling back to JCE implementation");
+            warningLog(e);
             return createDefaultEncryptedStream(output, keyMaterial, algorithm);
         }
     }
 
+    /** To avoid spamming logs with exceptions, we only log the exception at most once per minute. */
+    private static void warningLog(IOException exception) {
+        String message = "Unable to initialize cipher with OpenSSL, falling back to JCE implementation "
+                + "- see github.com/palantir/hadoop-crypto";
```
ellisjoe, line 104: Can we just log `OpenSSL not found or incompatible version - to fix see github.com/palantir/hadoop-crypto` and not even worry about logging the exception every 60 seconds?

sjrand: I think we'll want to be able to see the exception on the day when it doesn't work for some new reason, e.g., some new incompatibility with commons-crypto when we upgrade OpenSSL past 1.1.

(I was a fan of the "log the stacktrace at debug" idea, btw -- the once a minute thing is still gonna be quite noisy.)

iamdanfox: > not even worry about logging the exception every 60 seconds

I really want a way to be able to see this exception, even just to validate that they're all the same across the fleet. We currently _think_ they're all the same, but can't easily validate this.

> the "log the stacktrace at debug" idea

The reason I don't like debug is that when I opened up Aries, there were >20 services hitting this problem, and I'm not going to spend time to go through and open a bunch of cdconfig PRs to dial them up to debug. I want to be able to quickly assess whether we're hitting something new, or this is the same old problem we already know how to solve.

> once a minute thing is still gonna be quite noisy

Is once a minute really going to be that noisy? Looking at my sls-logging dashboard, I can see a lot of services log >200lines per second... going from 12,000 -> 12,001 per minute seems like not that bad? Alternatively I can happily change this to log the exception just the first time it hits, assuming it'll be the same stacktrace for the lifetime of the service?

sjrand: Re: the once a minute idea being noisy, it's just that stacktraces are big, so the log bloat is exaggerated by that.

Logging the exception only the first time it hits seems legit to me.

ellisjoe: I'm also cool with logging it the first time.

iamdanfox: Ok have updated

ellisjoe (approved):

PR #332, opened 2020-02-13
Don't initialize the FS twice

leonz:
## Before this PR
`FileSystem.get` [already calls initialize()](https://github.com/apache/hadoop/blob/rel/release-3.2.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L3303) within `createFileSystem()`.

Initializing twice leads to memory leaks since anything that is allocated within the first `initialize()` call cannot be closed.

## After this PR
==COMMIT_MSG==
Don't initialize the backing file system twice.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-332.v2.yml
+++ b/changelog/@unreleased/pr-332.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Don't initialize the backing file system twice.
+  links:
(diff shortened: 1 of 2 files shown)


ellisjoe (approved):

ferozco:
👍

PR #344, opened 2020-02-19
make readme link more explicit

tpetracca:
PDS-111703

I could see why this is brittle, but the current URL isn't enough (given empirical evidence).


crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
```diff
@@ -39,7 +39,7 @@
     private static final Properties PROPS = ApacheCiphers.forceOpenSsl(new Properties());
     private static final String AES_ALGORITHM = "AES/CTR/NoPadding";
     private static final String OPEN_SSL_INIT_WARNING = "Unable to initialize cipher with OpenSSL, falling back to "
-            + "JCE implementation - see github.com/palantir/hadoop-crypto";
+            + "JCE implementation - see github.com/palantir/hadoop-crypto#logwarn-lines-from-cryptostreamfactory";
```
iamdanfox: thoughts on just using `https://github.com/palantir/hadoop-crypto#faq` instead? I think we might unintentionally break this link by rewording that heading, but faq seems hard to get wrong

ellisjoe (approved):

PR #561, opened 2021-09-02
Reuse a buffer in ACDSI.InputAdapter

JacekLach:
## Before this PR

The allocation in .read was 30% of all allocations in internal product
during a period where we were allocating and GCing ~20GB of byte[]s
per minute.

## After this PR

==COMMIT_MSG==
Reduce byte[] allocations during reads via ApacheCtrDecryptingSeekableInput
==COMMIT_MSG==

### DIFF
--- a/crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
+++ b/crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
@@ -112,6 +112,7 @@ public final class ApacheCtrDecryptingSeekableInput extends CtrCryptoInputStream
 
         @Override
         public void close() throws IOException {
+            readBuffer = null;
             input.close();
         }
 
--- a/crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
+++ b/crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
@@ -16,6 +16,7 @@
 
 package com.palantir.crypto2.io;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.palantir.crypto2.cipher.ApacheCiphers;
 import com.palantir.crypto2.keys.KeyMaterial;
 import com.palantir.seekio.SeekableInput;
@@ -66,11 +67,12 @@ public final class ApacheCtrDecryptingSeekableInput extends CtrCryptoInputStream
         super.close();
     }
 
-    private static final class InputAdapter implements Input {
+    @VisibleForTesting
+    static final class InputAdapter implements Input {
         private final SeekableInput input;
         private final byte[] readBuffer = new byte[BUFFER_SIZE];
 
-        private InputAdapter(SeekableInput input) {
+        InputAdapter(SeekableInput input) {
             this.input = input;
         }
 
@@ -90,7 +92,12 @@ public final class ApacheCtrDecryptingSeekableInput extends CtrCryptoInputStream
                 int read = input.read(readBuffer, 0, chunk);
 
                 if (read == -1) {
-                    return totalRead;
+                    if (totalRead == 0) {
+                        // first read hit EOF
+                        return -1;
+                    } else {
+                        return totalRead;
+                    }
                 } else {
                     dst.put(readBuffer, 0, read);
                     totalRead += read;
--- a/crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
+++ b/crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
@@ -117,30 +117,18 @@ public final class ApacheCtrDecryptingSeekableInput extends CtrCryptoInputStream
         }
 
         private void resize(int required) {
-            int size = readBuffer.length;
-            readBuffer = new byte[newLength(size, required - size, size << 1)];
+            readBuffer = new byte[newLength(readBuffer.length, required)];
         }
 
-        // copied from jdk.internal.util.ArraysSupport
+        // some jvms reserve header words in an array, reducing its max size
         private static final int MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
 
-        private static int newLength(int oldLength, int minGrowth, int prefGrowth) {
-            // assert oldLength >= 0
-            // assert minGrowth > 0
-
-            int newLength = Math.max(minGrowth, prefGrowth) + oldLength;
+        private static int newLength(int oldLength, int targetLength) {
+            int newLength = Math.max(targetLength, oldLength << 1);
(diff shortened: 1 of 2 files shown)


JacekLach:
@ellisjoe I can't assign reviewers, would appreciate a look

JacekLach:
image
jfr screenshot for this being a significant issue

JacekLach:
I considered checking if `buffer.hasArray()` and if so writing directly to the array - however on this code path the byte buffer is always allocated via `ByteBuffer.allocateDirect(this.bufferSize + cipher.getBlockSize());`

https://github.com/apache/commons-crypto/blob/baa0d8fd73ee4756f1ae397afbdce8db0a9a2580/src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java#L203-L207

therefore the buffer will never have an array we can write to, and we have to pay for this copy

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -110,8 +112,38 @@ public void seek(long position) throws IOException {
...
             input.close();
         }
-    }
 
+        private void resize(int required) {
+            int size = readBuffer.length;
+            readBuffer = new byte[newLength(size, required - size, size << 1)];
+        }
+
+        // copied from jdk.internal.util.ArraysSupport
```
carterkozak: Isn't that GPL?

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -82,11 +82,13 @@ public int read(long position, byte[] buffer, int offset, int length) throws IOE
 
         @Override
         public int read(ByteBuffer dst) throws IOException {
-            byte[] bytes = new byte[dst.remaining()];
-            int read = input.read(bytes, 0, bytes.length);
+            if (readBuffer.length < dst.remaining()) {
+                resize(dst.remaining());
```
carterkozak: Why is resizing necessary? Can't we read from the source in a loop using the buffer we already have?

JacekLach: in practice the buffer should always fit, since we read into a byte buffer created with same size in https://github.com/palantir/hadoop-crypto/blob/e42e0496f6a1efe0a7c6a94d81509406672c5f5d/crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java#L45

so this covers the case that something changes in the calling code to use a larger buffer. looping intuitively feels more expensive, more os calls vs single allocation that gets reused from then on (assuming buffer size stays constant)

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
JacekLach: bleh, yes it is. I'll rewrite & simplify a bit

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
carterkozak: > looping intuitively feels more expensive

Beyond 8kB (or in some tests 16kB) larger buffers don't tend to reduce overhead for network or local disk operations in my experience. I'd bias toward the loop unless there's a benchmark that proves larger values can work better (and in that case I'd update the buffer size while keeping the loop as a fallback).

JacekLach: Sure, I guess. The code gets more complicated but can do that

carterkozak (approved):
lgtm

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -67,8 +67,8 @@ public void close() throws IOException {
     }
 
     private static final class InputAdapter implements Input {
-
         private SeekableInput input;
+        private byte[] readBuffer = new byte[BUFFER_SIZE];
```
carterkozak: final, no need to null this out after close imo -- assuming we're not holding a reference to the closed stream for a long time.
```suggestion
        private final byte[] readBuffer = new byte[BUFFER_SIZE];
```

JacekLach: yep, makes sense if we know the buffer is constant size

JacekLach:
Can someone click the 'generate changelog' button, it's not available for me :P

JacekLach:
Hm, it timed out on the second push too, but I really don't see how that could be caused by this change. I'll try another run tomorrow, I guess, hoping that whatever's wrong atm will pass

crypto-core/src/main/java/com/palantir/crypto2/io/ApacheCtrDecryptingSeekableInput.java
```diff
@@ -82,14 +82,23 @@ public int read(long position, byte[] buffer, int offset, int length) throws IOE
...
-                dst.put(bytes, 0, read);
+            int toRead = dst.remaining();
+            int totalRead = 0;
+
+            while (toRead > 0) {
+                int chunk = Math.min(toRead, readBuffer.length);
+                int read = input.read(readBuffer, 0, chunk);
+
+                if (read == -1) {
+                    return totalRead;
```
ellisjoe: I think this needs to be:
```
if (read == -1) {
    if (totalRead == 0) {
        totalRead = -1;
    }
}
```

there's a reference implementation here: https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/stream/input/StreamInput.java#L58-L75

I think this logic ends up being complex enough to warrant a few tests as well. When I was playing with this myself I pulled the InputAdapter up. I'm still trying to find some common util that we can use here but haven't found anything yet.

JacekLach: I was under the impression that `DecryptionTests` would cover this - but in fact am struggling to have those run locally (tests in intellij die with 'Process finished with exit code 134 (interrupted by signal 6: SIGABRT)') - is there a trick to it?

JacekLach: d1e832eb4180f26033b1e6229dcb13f4be7276a5

JacekLach:
it does look like the circleci hang was related to that bug - unclear how but clearly I don't understand the mechanics of these tests

ellisjoe (approved):

PR #576, opened 2021-10-07
Backport buffer re-use (#561) to 2.x

dtobin:
## Before this PR
The 3.0.0 release includes a breaking upgrade to Hadoop 3.x. We want to backport the buffer re-use fix from #561 to 2.x so that consumers unable to upgrade their Hadoop version can pick up the fix.

## After this PR

==COMMIT_MSG==
Reduce byte[] allocations during reads via ApacheCtrDecryptingSeekableInput
==COMMIT_MSG==


dtobin:
@carterkozak @ellisjoe I don't have admin permissions on the repo to perform the autorelease - will need help with that. Let me know if there is a convention for release branch naming.

carterkozak (approved):

PR #586, opened 2021-11-08
Expand EncryptionBenchmark

carterkozak:
1. This adds an additional parameter which splits the input data
   into 16k writes rather than encoding the full buffer at once.
2. The encrypted bytes are returned from the benchmark to avoid
   clever dead-store optimizations. This piece may be replaced
   with a simple OutputStream wrapping the jmh Blackhole, but I
   don't expect the array allocation to make a substantial
   difference here.
3. Crypto streams are closed. Unclear what, if any, resources
   may be leaked.

### Initial results on Java 17 LTS (JDK 17.0.1, OpenJDK 64-Bit Server VM, 17.0.1+12-LTS)

```
Benchmark                          (numBytes)  (writeStrategy)   Mode  Cnt     Score      Error  Units
EncryptionBenchmark.apacheEncrypt     1048576    ENTIRE_BUFFER  thrpt    4  1482.463 ±  126.144  ops/s
EncryptionBenchmark.apacheEncrypt     1048576          CHUNKED  thrpt    4  1487.755 ±  156.773  ops/s
EncryptionBenchmark.apacheEncrypt    10485760    ENTIRE_BUFFER  thrpt    4   103.725 ±    9.012  ops/s
EncryptionBenchmark.apacheEncrypt    10485760          CHUNKED  thrpt    4   107.081 ±    6.937  ops/s
EncryptionBenchmark.apacheEncrypt   104857600    ENTIRE_BUFFER  thrpt    4    11.268 ±    0.740  ops/s
EncryptionBenchmark.apacheEncrypt   104857600          CHUNKED  thrpt    4    10.643 ±    0.193  ops/s
EncryptionBenchmark.ctrEncrypt        1048576    ENTIRE_BUFFER  thrpt    4   731.421 ± 4031.046  ops/s
EncryptionBenchmark.ctrEncrypt        1048576          CHUNKED  thrpt    4  1548.771 ±  117.477  ops/s
EncryptionBenchmark.ctrEncrypt       10485760    ENTIRE_BUFFER  thrpt    4    29.861 ±    2.551  ops/s
EncryptionBenchmark.ctrEncrypt       10485760          CHUNKED  thrpt    4   116.432 ±   13.402  ops/s
EncryptionBenchmark.ctrEncrypt      104857600    ENTIRE_BUFFER  thrpt    4     2.865 ±    0.173  ops/s
EncryptionBenchmark.ctrEncrypt      104857600          CHUNKED  thrpt    4    11.809 ±    1.016  ops/s
[...]

PR #587, opened 2021-11-09
CryptoStreamFactory creates sympathetic chunking OutputStreams

carterkozak:
For detailed analysis of the problem, see
https://github.com/palantir/hadoop-crypto/pull/586

CipherOutputStream appears to perform _very_ poorly on large
buffers, but substantially better when data is segmented into
small chunks which can be done by looping over the original buffer.

With this in place, the openssl wrappeer no longer provides any
benefit over JCE.

==COMMIT_MSG==
CryptoStreamFactory creates sympathetic chunking OutputStreams with performance characteristics _matching_ the apache commons-crypto implementation
==COMMIT_MSG==

### DIFF
--- a/crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
+++ b/crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
@@ -22,6 +22,7 @@ import com.palantir.crypto2.cipher.SeekableCipher;
 import com.palantir.crypto2.cipher.SeekableCipherFactory;
 import com.palantir.crypto2.keys.KeyMaterial;
 import com.palantir.seekio.SeekableInput;
+import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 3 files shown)


crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
```diff
@@ -150,4 +151,53 @@ public void close() throws IOException {
             input.close();
         }
     }
+
+    /**
+     * {@link ChunkingOutputStream} limits the size of individual writes to the wrapped {@link OutputStream}
+     * in order to prevent degraded performance on large buffers as described in
+     * <a href="https://github.com/palantir/hadoop-crypto/pull/586">hadoop-crypto#586</a>.
+     */
+    static final class ChunkingOutputStream extends FilterOutputStream {
```
ellisjoe, line 160: any reason a `BufferedOutputStream` won't work for you here?

carterkozak: Yep! BufferedOutputStream actually does the opposite of what we want here! Given a large input (beyond the configured buffer size), the BufferedOutputStream will flush any data in its buffer, and write the input buffer directly to the delegate stream as is.

In most cases a BufferedOutputStream is helpful to avoid native overhead for crypto, but I think that's a bit outside of the scope of this change, and introducing a buffered stream around our chunking stream may result in additional unnecessary copies.

ellisjoe (approved):

PR #592, opened 2021-11-16
CryptoStreamFactory logs reasonably when OpenSSL is or isn't detected

carterkozak:
Logging occurs once, the first time an exception is hit. The
message and stack trace are logged at `INFO` level, and
functionality continues as expected.

==COMMIT_MSG==
CryptoStreamFactory logs reasonably when OpenSSL is or isn't detected
==COMMIT_MSG==

### DIFF
--- a/crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
+++ b/crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
@@ -41,10 +41,6 @@ public final class CryptoStreamFactory {
     private static final Logger log = LoggerFactory.getLogger(CryptoStreamFactory.class);
     private static final Properties PROPS = ApacheCiphers.forceOpenSsl(new Properties());
     private static final String AES_ALGORITHM = "AES/CTR/NoPadding";
-    private static final String OPEN_SSL_INIT_WARNING = "Unable to initialize cipher with OpenSSL, falling back to "
-            + "JCE implementation - see github.com/palantir/hadoop-crypto#faq";
-
-    private static volatile boolean fullExceptionLoggedAlready = false;
 
     private static final Supplier<Boolean> OPENSSL_IS_AVAILABLE = Suppliers.memoize(() -> {
         try {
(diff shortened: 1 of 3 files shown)


crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
```diff
@@ -44,14 +46,25 @@
 
     private static volatile boolean fullExceptionLoggedAlready = false;
```
carterkozak: I wrote this as an entirely additive change. We could get rid of the `fullExceptionLoggedAlready` flag, and the catch+fallback case entirely, but there may be cases that helps us which I'm not aware of, so I opted to minimize risk.

ellisjoe: I'm good with killing consolidating the exception handling. There aren't any other cases I'm aware of where cipher construction can throw so can also probably remove the try catch.

crypto-core/src/main/java/com/palantir/crypto2/io/CryptoStreamFactory.java
```diff
@@ -83,7 +96,7 @@ public static InputStream decrypt(InputStream input, KeyMaterial keyMaterial, St
      * cipher {@code algorithm}. When OpenSSL is available an implementation that uses AES-NI will be returned.
      */
     public static OutputStream encrypt(OutputStream output, KeyMaterial keyMaterial, String algorithm) {
-        return encrypt(output, keyMaterial, algorithm, false);
+        return encrypt(output, keyMaterial, algorithm, !OPENSSL_IS_AVAILABLE.get());
```
ellisjoe: I think things may read a bit better if we rewrite the encrypt method below like:
```
static OutputStream encrypt(OutputStream output, KeyMaterial keyMaterial, String algorithm, boolean forceJce) {
    if (algorithm.equals(AES_ALGORITHM) && OPENSSL_IS_AVAILABLE.get() && !forceJce) {
        return createApacheEncryptedStream(output, keyMaterial);
    } else {
        return createDefaultEncryptedStream(output, keyMaterial, algorithm);
    }
```

I also don't remember which tests needed the `forceJce` flag so may be able to consider removing that completely now as well.

carterkozak: Removal would make it impossible to test both paths on the same CI run, so I don't think we should do that unless we're removing one of the two paths entirely :-)
Let's hold off cleaning up this codebase for an isolated commit rather than coupling it with logging cleanup, I'll add a minimal refactor to retain the `false` arg for the time being.

PR #596, opened 2022-01-06
Circle uses java-library-oss template

schlosna:
## Before this PR

Circle CI config was unmanaged, blocking many excavators using JDK 11+ gradle plugins

## After this PR

==COMMIT_MSG==
Circle uses java-library-oss template
==COMMIT_MSG==

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,94 +2,175 @@
 # To request a modification to the general template, file an issue on Excavator.
 # To manually manage the CircleCI configuration for this project, remove the .circleci/template.sh file.
 
-version: 2
+version: 2.1
 jobs:
-
   compile:
-    docker: [{ image: 'ellisjoe/openjdk-java-openssl:java-8' }]
-    resource_class: xlarge
+    docker: [{ image: 'cimg/openjdk:11.0.10-node' }]
+    resource_class: large
     environment:
-      GRADLE_OPTS: -Dorg.gradle.console=plain -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false
-
+      CIRCLE_TEST_REPORTS: /home/circleci/junit
+      CIRCLE_ARTIFACTS: /home/circleci/artifacts
+      GRADLE_OPTS: -Dorg.gradle.jvmargs='-Xmx2g' -Dorg.gradle.workers.max=2
+      _JAVA_OPTIONS: -XX:ActiveProcessorCount=4 -XX:MaxRAM=8g -XX:ErrorFile=/home/circleci/artifacts/hs_err_pid%p.log -XX:HeapDumpPath=/home/circleci/artifacts
     steps:
       - checkout
-      - restore_cache: { key: 'gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}' }
-      - restore_cache: { key: 'gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}' }
-      - run: ./gradlew --no-daemon --parallel --continue classes testClasses
+      - run:
+          name: delete_unrelated_tags
+          command: |
+            ALL_TAGS=$(git tag --points-at HEAD)
+
+            if [ -z "$ALL_TAGS" ]; then
+                echo "No-op as there are no tags on the current commit ($(git rev-parse HEAD))"
+                exit 0
+            fi
+
+            if [ -z "${CIRCLE_TAG:+x}" ]; then
+                echo "Non-tag build, deleting all tags which point to HEAD: [${ALL_TAGS/$'\n'/,}]"
+                echo "$ALL_TAGS" | while read -r TAG; do git tag -d "$TAG" 1>/dev/null; done
+                exit 0
+            fi
+
+            TAGS_TO_DELETE=$(echo "$ALL_TAGS" | grep -v "^$CIRCLE_TAG$" || :)
+            if [ -z "$TAGS_TO_DELETE" ]; then
+                echo "No-op as exactly one tag ($CIRCLE_TAG) points to HEAD"
+                exit 0
+            fi
+
+            echo "Detected tag build, deleting all tags except '$CIRCLE_TAG' which point to HEAD: [${TAGS_TO_DELETE/$'\n'/,}]"
+            echo "$TAGS_TO_DELETE" | while read -r TAG; do git tag -d "$TAG" 1>/dev/null; done
+      - restore_cache: { key: 'gradle-wrapper-v2-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}' }
+      - restore_cache: { key: 'compile-gradle-cache-v2-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}' }
+      - run: ./gradlew --parallel --stacktrace classes testClasses
       - save_cache:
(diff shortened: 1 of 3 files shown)


carterkozak:
I don't think this image has `/usr/lib/libcrypto.so`.  we could fall back to the result of something like `strings /etc/ld.so.cache | grep '/libcrypto.so$'`, using whichever version is available from the container, but I don't know the history behind the `/usr/lib/libcrypto.so` path.

.circleci/config.yml
```diff
@@ -2,94 +2,186 @@
...
-      - run: ./gradlew --no-daemon --parallel --continue classes testClasses
+      - run:
+          name: install_openssl
+          command: |
+            sudo apt-get update -q
+            sudo apt-get install -qy openssl
+            sudo ln -s /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1 /usr/lib/ssl/libcrypto.so
+            sudo ln -s /usr/lib/x86_64-linux-gnu/libssl.so.1.1 /usr/lib/ssl/libssl.so
+            sudo ln -s /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1 /usr/lib/libcrypto.so
+            sudo ln -s /usr/lib/x86_64-linux-gnu/libssl.so.1.1 /usr/lib/libssl.so
```
carterkozak: Should this move to the `unit-test:` section?

schlosna:
I'm mainly interested in unblocking the excavators and ensuring things can build on recent JDK 11+ and gradle tooling while I had some cycles. Looks like it may be easier to hold off on this until https://github.com/palantir/hadoop-crypto/pull/588 merges

> I don't think this image has `/usr/lib/libcrypto.so`. we could fall back to the result of something like `strings /etc/ld.so.cache | grep '/libcrypto.so$'`, using whichever version is available from the container, but I don't know the history behind the `/usr/lib/libcrypto.so` path.

Yeah, these are currently using @ellisjoe 's `ellisjoe/ibm-java-openssl:java-8` container images which already have openssl installed per https://github.com/ellisjoe/openjdk-java-openssl/blob/develop/Dockerfile , but the CircleCI openjdk convenience images do not currently have openssl installed.

carterkozak:
Still have investigation in progress before #588 can be merged, but it currently results in a great deal of noise until something like #592 is released.

.circleci/config.yml
```diff
@@ -2,94 +2,183 @@
...
+      - restore_cache: { key: 'gradle-wrapper-v2-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}' }
+      - restore_cache: { key: 'unit-test-gradle-cache-v2-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}' }
+      - run:
+          name: link_openssl
+          command: |
+            sudo apt-get update -q && sudo apt-get install -qy libssl-dev
+            sudo ln -s /lib/x86_64-linux-gnu/libcrypt.so /usr/lib/libcrypto.so
+            sudo ln -s /lib/x86_64-linux-gnu/libcrypt.so /usr/lib64/libcrypto.so
+            echo "Linked openssl /usr/lib/libcrypto.so"
+            ls -al /lib/libcrypto* /usr/lib/libcrypto* /usr/lib64/libcrypto* /usr/lib/x86_64-linux-gnu/libcrypto*
```
schlosna, line 97: This is the section we'll need to figure out how we want to fold into circle template so that it doesn't immediately get removed by excavator though unit tests should block that from merging, but that also means the excavator won't be updating everything else including CI images.

For future reference, see https://issues.apache.org/jira/browse/HADOOP-12845 & https://issues.apache.org/jira/browse/HADOOP-11216

Resulting libcrypto symlinks (we need `/usr/lib64/libcrypto.so`)
```
lrwxrwxrwx 1 root root      33 Jan  6 20:36 /lib/libcrypto.so -> /lib/x86_64-linux-gnu/libcrypt.so
lrwxrwxrwx 1 root root      33 Jan  6 20:36 /usr/lib64/libcrypto.so -> /lib/x86_64-linux-gnu/libcrypt.so
lrwxrwxrwx 1 root root      33 Jan  6 20:36 /usr/lib/libcrypto.so -> /lib/x86_64-linux-gnu/libcrypt.so
-rw-r--r-- 1 root root 5541876 Nov 24 13:20 /usr/lib/x86_64-linux-gnu/libcrypto.a
lrwxrwxrwx 1 root root      16 Nov 24 13:20 /usr/lib/x86_64-linux-gnu/libcrypto.so -> libcrypto.so.1.1
-rw-r--r-- 1 root root 2954080 Nov 24 13:20 /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1
```

schlosna (commented):
@carterkozak builds are green now, but would you prefer I hold off on this until after #592 & #588 merge?

.circleci/config.yml
```diff
@@ -2,94 +2,183 @@
...
+          command: git status --porcelain
+          when: always
+      - save_cache:
+          key: 'publish-gradle-cache-v2-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}'
+          paths: [ ~/.gradle/caches ]
+      - store_test_results: { path: ~/junit }
+      - store_artifacts: { path: ~/artifacts }
 
+  markdown:
+    docker: [{ image: 'raviqqe/liche:0.1.1' }]
```
carterkozak, line 154: I thought we'd opted out of this check by default in our templates due to the issues with links to other github projects?

carterkozak:
@schlosna Ordering doesn't make any difference to me, doesn't impact the other PRs that are up and gives us more modern testing.

Was there a reason we tested on ibm jdks? Is that something we need to retain?

.circleci/config.yml
```diff
@@ -2,94 +2,183 @@
...
-
+      CIRCLE_TEST_REPORTS: /home/circleci/junit
+      CIRCLE_ARTIFACTS: /home/circleci/artifacts
+      GRADLE_OPTS: -Dorg.gradle.jvmargs='-Xmx2g' -Dorg.gradle.workers.max=2
+      _JAVA_OPTIONS: -XX:ActiveProcessorCount=4 -XX:MaxRAM=8g -XX:ErrorFile=/home/circleci/artifacts/hs_err_pid%p.log -XX:HeapDumpPath=/home/circleci/artifacts
     steps:
-      - checkout
-      - restore_cache: { key: 'gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}' }
-      - restore_cache: { key: 'gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}' }
-      - run: ./gradlew --no-daemon --parallel --continue crypto-core:test
```
schlosna, line 53: @carterkozak Some consumers of hadoop-crypto may use J9 for JVM runtime so this was running the tests using IBM's J9 JVM via https://github.com/ellisjoe/ibm-java-openssl/blob/develop/Dockerfile 

Defer to you & @ellisjoe if you'd want to keep these J9 CI tests

schlosna:
Before we merge this, we'll need to adjust the Github expected checks:

* Add Required
    * `ci/circleci: check`
    * `ci/circleci: trial-publish`

* Remove Required
    * `ci/circleci: build`
    * `ci/circleci: build-ibm`
    * `ci/circleci: publish`

schlosna:
closing as this was updated via #619 . will rerun excavator

PR #619, opened 2022-09-30
Use CircleCI template with JDK 11

schlosna:
## Before this PR
The circleci setup was using a custom JDK 8 container image with openssl added to enable testing with Hadoop native library crypto. Unfortunately, the custom circle config is causing a bunch of stuck excavators due to Gradle infrastructure requiring JDK 11.

See:

- https://github.com/ellisjoe/openjdk-java-openssl/blob/develop/Dockerfile
- https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/NativeLibraries.html
- https://issues.apache.org/jira/browse/HADOOP-12845

## After this PR
==COMMIT_MSG==
Use CircleCI template with JDK 11
==COMMIT_MSG==

## Possible downsides?

Does not setup & test IBM JDK yet

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,94 +2,145 @@
 # To request a modification to the general template, file an issue on Excavator.
 # To manually manage the CircleCI configuration for this project, remove the .circleci/template.sh file.
 
-version: 2
+version: 2.1
 jobs:
 
-  compile:
-    docker: [{ image: 'ellisjoe/openjdk-java-openssl:java-8' }]
-    resource_class: xlarge
+  check:
+    docker: [{ image: 'cimg/openjdk:11.0.10-node' }]
+    resource_class: large
     environment:
-      GRADLE_OPTS: -Dorg.gradle.console=plain -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false
-
+      CIRCLE_TEST_REPORTS: /home/circleci/junit
+      CIRCLE_ARTIFACTS: /home/circleci/artifacts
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -2,94 +2,145 @@
...
-
+      CIRCLE_TEST_REPORTS: /home/circleci/junit
+      CIRCLE_ARTIFACTS: /home/circleci/artifacts
+      GRADLE_OPTS: -Dorg.gradle.workers.max=1 -Dorg.gradle.jvmargs='-Xmx2g --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED'
+      _JAVA_OPTIONS: -XX:ActiveProcessorCount=2 -XX:MaxRAM=4g -XX:ErrorFile=/home/circleci/artifacts/hs_err_pid%p.log -XX:HeapDumpPath=/home/circleci/artifacts
     steps:
       - checkout
-      - restore_cache: { key: 'gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}' }
-      - restore_cache: { key: 'gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}' }
-      - run: ./gradlew --no-daemon --parallel --continue crypto-core:test
```
schlosna, line 53: Note that we previously tested against IBM JDK 8 using the container image from https://github.com/ellisjoe/ibm-java-openssl/blob/develop/Dockerfile, but that  has not been migrated over. We could potentially `sudo apt install adoptopenjdk-11-openj9` and test against that, but would prefer to defer that for another day.

ellisjoe: I'm fine no longer testing against the IBM JVM. I'm not aware of any usage and the overhead of maintaining that additional test has been quite high.

schlosna:
once this is approved, we'll need to adjust the required checks for repo to `ci/circleci: check` and `ci/circleci: trial-publish` and remove the other old tasks

carterkozak (approved):
lgtm, thank you :-)

schlosna:
🚀

PR #621, opened 2022-10-03
Migrate to JUnit 5

schlosna:
## Before this PR
Project was using JUnit 4 and JUnit 5 excavator was failing.

## After this PR

==COMMIT_MSG==
Migrate to JUnit 5
==COMMIT_MSG==


carterkozak (approved):
Confirmed 184 tests in circleci before and after this change 👍
Thanks!

PR #629, opened 2022-10-24
Fix JMH dependencies

carterkozak:
==COMMIT_MSG==
Fix JMH dependencies
==COMMIT_MSG==


crypto-core/build.gradle
```diff
@@ -11,8 +11,8 @@ dependencies {
     implementation 'com.palantir.safe-logging:preconditions'
     implementation 'com.palantir.safe-logging:safe-logging'
 
-    annotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess'
-    compileOnly 'org.openjdk.jmh:jmh-generator-annprocess'
+    testAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess'
+    testCompileOnly 'org.openjdk.jmh:jmh-generator-annprocess'
```
carterkozak, line 15: The benchmark lives in `src/test/java`

schlosna: maybe separate, but should we move them to `src/jmh/java` use use [jmh-gradle-plugin](https://github.com/melix/jmh-gradle-plugin)?

schlosna (commented):
Approving as this is strictly an improvement, though wondering if we should move the benchmark

schlosna (approved):

crypto-core/build.gradle
carterkozak: Agree that would be cleaner, I've had some dependency issues with the gradle plugin that are probably solved or irrelevant at this point -- I usually create a separate module instead, but either way would be an improvement over reusing test sources.

PR #636, opened 2022-12-01
Throw if JDK-8292158 may cause AES-CTR encryption corruption

schlosna:
## Before this PR
JDK bug https://bugs.openjdk.org/browse/JDK-8292158 could cause corruption of AES-CTR streams when running on CPUs with AVX-512 vectorized AES support (e.g. Intel Ice Lake, as used in AWS 6th generation EC2 instances).

This issue should be resolved in OpenJDK 11.0.18, 15.0.10, 17.0.6, 19.0.2, and 20.0.0 when released.

One can temporarily workaround this issue be adding one of the following sets of JVM arguments:

* `-XX:+UnlockDiagnosticVMOptions -XX:-UseAESCTRIntrinsics` to disable only AES-CTR intrinsics
* `-XX:+UnlockDiagnosticVMOptions -XX:-UseAESIntrinsics` to disable all AES intrinsics
* `-XX:-UseAES` to disable all AES native instructions
* `-XX:UseAVX=2` to disable all AVX-512 instructions

See https://github.com/palantir/sls-packaging/pull/1411

## After this PR
==COMMIT_MSG==
To avoid potential stream corruption, throw if JDK-8292158 may cause AES-CTR encryption corruption.

Determine if JVM is impacted by JDK-8292158 which can corrupt AES-CTR encryption streams. This bug impacts JDKs up to 11.0.18, 15.0.10, 17.0.6, 19.0.2 and when running on CPUs with AVX-512 vectorized AES support.

See https://bugs.openjdk.org/browse/JDK-8292158 introduced by https://bugs.openjdk.org/browse/JDK-8233741
==COMMIT_MSG==

## Possible downsides?
This will fail fast and trigger runtime failures when running on un-mitigated JVMs rather than possible data corruption.

### DIFF
--- a/changelog/@unreleased/pr-636.v2.yml
+++ b/changelog/@unreleased/pr-636.v2.yml
@@ -1,0 +1,10 @@
+type: improvement
+improvement:
+  description: |-
+    To avoid potential stream corruption, throw if JDK-8292158 may cause AES-CTR encryption corruption.
+
+    Determine if JVM is impacted by JDK-8292158 which can corrupt AES-CTR encryption streams. This bug impacts JDKs up to 11.0.18, 15.0.10, 17.0.6, 19.0.2 and when running on CPUs with AVX-512 vectorized AES support.
+
+    See https://bugs.openjdk.org/browse/JDK-8292158 introduced by https://bugs.openjdk.org/browse/JDK-8233741
+  links:
+  - https://github.com/palantir/hadoop-crypto/pull/636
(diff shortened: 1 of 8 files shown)


carterkozak:
Could we run the failing test case instead of attempting to determine whether or not the jvm+hardware combination is impacted?

carterkozak:
CI tests will be difficult/impossible, but a manual verification should be sufficient.

schlosna:
Manually tested this on Intel Ice Lake and it identifies and catches the corruption

PR #638, opened 2022-12-03
PathConvertingFileSystem converts FileStatus objects in-place

rshkv:
## Before this PR
The current implementation of `PathConvertingFileSystem#toReturnFileStatus`, the method that converts file statuses from the delegating file system, creates new `FileStatus` objects. This doesn't work with custom implementation of `FileStatus`.

## After this PR
Convert file statuses from the delegating file status by setting the new path instead over creating a whole new object. I couldn't find an implementation of `FileStatus` that would be negatively affected by this. To the contrary: if file system implementations return custom file status implementation, the delegating file system should respect those and not return new statuses of a different type.

==COMMIT_MSG==
PathConvertingFileSystem renames paths without creating new FileStatus objects.
==COMMIT_MSG==

### DIFF
--- a/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/PathConvertingFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/PathConvertingFileSystem.java
@@ -154,19 +154,8 @@ public final class PathConvertingFileSystem extends DelegatingFileSystem {
         return fromFunc.apply(path);
     }
 
-    private FileStatus toReturnFileStatus(FileStatus status) throws IOException {
-        // same as FileStatus copy constructor
-        return new FileStatus(
-                status.getLen(),
-                status.isDirectory(),
-                status.getReplication(),
-                status.getBlockSize(),
-                status.getModificationTime(),
-                status.getAccessTime(),
-                status.getPermission(),
-                status.getOwner(),
-                status.getGroup(),
-                status.isSymlink() ? status.getSymlink() : null, // getSymlink throws if file is not a symlink
-                from(status.getPath()));
+    private FileStatus toReturnFileStatus(FileStatus status) {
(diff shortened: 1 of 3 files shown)


hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/PathConvertingFileSystem.java
```diff
@@ -154,19 +154,8 @@ private Path from(Path path) {
...
-                status.getAccessTime(),
-                status.getPermission(),
-                status.getOwner(),
-                status.getGroup(),
-                status.isSymlink() ? status.getSymlink() : null, // getSymlink throws if file is not a symlink
-                from(status.getPath()));
+    private FileStatus toReturnFileStatus(FileStatus status) {
+        status.setPath(from(status.getPath()));
+        return status;
     }
```
schlosna, line 160: Is there a test case that covers the behavior we want here?

rshkv: Yeah, PathConvertingFileSystemTest covers the conversion here: https://github.com/palantir/hadoop-crypto/blob/d0853ab96feca47a7e6f385f52bc71d8a2a54357/hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/PathConvertingFileSystemTest.java#L121-L135

rshkv: Also testing e.g. the object reference felt too much like an implementation detail. But can add.

rshkv: On further thought testing the new behavior makes sense: added in 410ff78.

schlosna (approved):

PR #667, opened 2023-03-01
Do not set scheme if unset

mswintermeyer:
## Before this PR
StandaloneEncryptedFileSystem#setUriSchemeFunc will always set the uri's scheme, even if it's not already set.

## After this PR
We only set the scheme if it's already set. This matches the behavior of this class in 2.9.2: https://github.com/palantir/hadoop-crypto/pull/442/files#diff-050a60db63beecf42d4d99e5e9132ee7f6dc68ae9b552e37ff5ccd2a109806d6L166
==COMMIT_MSG==
StandaloneEncryptedFileSystem returns to its old behavior, of only overriding the scheme when the uri uses a scheme
==COMMIT_MSG==

### DIFF
--- a/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
@@ -164,6 +164,9 @@ public final class StandaloneEncryptedFileSystem extends FilterFileSystem {
 
     private static Function<URI, URI> setUriSchemeFunc(final String scheme) {
         return uri -> {
+            if (uri.getScheme() == null) {
+                return uri;
+            }
             try {
                 return new URI(
                         scheme,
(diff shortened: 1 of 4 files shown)


iambenjackson (approved):

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/StandaloneEncryptedFileSystem.java
```diff
@@ -164,6 +164,9 @@ private static Function<Path, Path> setSchemeFunc(final String scheme) {
 
     private static Function<URI, URI> setUriSchemeFunc(final String scheme) {
         return uri -> {
+            if (uri.getScheme() == null) {
```
ellisjoe, line 167: Two quick things:
- Do you have an example where things have broken when this scheme is set?
- Can we write a test that illustrates this behavior in action?

alexmaass (approved):

ellisjoe (approved):

PR #696, opened 2023-05-12
Fix: allow resuming partially deleted encrypted files

mswintermeyer:
## Before this PR
If a process is interrupted midway through EncryptedFileSystem#delete, we could delete the keymaterial file without deleting the actual file, and successive deletes could be unsuccessful.

`EncryptedFileSystemTest#testDelete_keyMaterialAlreadyDeleted` already attempted to test this, but it was not a representative test because `InMemoryKeyStorageStrategy` never threw errors.

## After this PR

==COMMIT_MSG==
EncryptedFileSystem#delete now correctly can delete a partially deleted (aka only keymaterial deleted) file.
==COMMIT_MSG==

### DIFF
--- a/crypto-keys/src/main/java/com/palantir/crypto2/keys/KeyStorageStrategy.java
+++ b/crypto-keys/src/main/java/com/palantir/crypto2/keys/KeyStorageStrategy.java
@@ -16,6 +16,8 @@
 
 package com.palantir.crypto2.keys;
 
+import java.io.IOException;
+
 /**
  * The strategy used to store the per file {@link KeyMaterial} used for encryption operations.
  */
@@ -24,15 +26,15 @@ public interface KeyStorageStrategy {
     /**
      * Stores the {@link KeyMaterial} for a file with the given {@code fileKey}.
      */
-    void put(String fileKey, KeyMaterial keyMaterial);
+    void put(String fileKey, KeyMaterial keyMaterial) throws IOException;
 
     /**
      * Retrieves the {@link KeyMaterial} for a file with the given {@code fileKey}.
      */
-    KeyMaterial get(String fileKey);
+    KeyMaterial get(String fileKey) throws IOException;
 
     /**
      * Removes the {@link KeyMaterial} for a file with the given {@code fileKey}.
      */
-    void remove(String fileKey);
+    void remove(String fileKey) throws IOException;
 }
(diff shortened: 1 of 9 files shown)


hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -171,7 +171,9 @@ public boolean delete(Path path, boolean recursive) throws IOException {
         if (recursive) {
             throw new UnsupportedOperationException("EncryptedFileSystem does not support recursive deletes");
         }
-        keyStore.remove(path.toString());
+        // If delete fails partway through, after deleting key but not file, make it possible to fully delete
+        // the file on a subsequent call.
+        tryRemoveKey(path);
```
mswintermeyer, line 176: Beyond this change, Nimai suggested reordering this so that we delete the file before we delete the keymaterial. Since leaving behind a useless keymaterial file on partial-delete is perhaps less confusing than leaving behind an undecryptable real file that's unclear if it's important or not

crypto-keys/src/main/java/com/palantir/crypto2/keys/KeyStorageStrategy.java
```diff
@@ -24,15 +26,15 @@ public interface KeyStorageStrategy {
     /**
      * Stores the {@link KeyMaterial} for a file with the given {@code fileKey}.
      */
-    void put(String fileKey, KeyMaterial keyMaterial);
+    void put(String fileKey, KeyMaterial keyMaterial) throws IOException;
```
ellisjoe: I'd really like to avoid throwing IOExceptions to manage flow control. It seems like `remove` is the only problematic method here, so it might be worth adding a new method that differentiates between NotFound, Failed, and Succeeded?

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/InMemoryKeyStorageStrategy.java
```diff
@@ -28,17 +30,21 @@ public final class InMemoryKeyStorageStrategy implements KeyStorageStrategy {
...
+    public KeyMaterial get(String fileKey) throws IOException {
         return store.get(fileKey);
     }
 
     @Override
-    public void remove(String fileKey) {
+    public void remove(String fileKey) throws IOException {
+        if (!store.containsKey(fileKey)) {
+            // Mimic what the non-test-in-memory version would do if trying to remove it
+            throw new NoSuchFileException(fileKey);
```
ellisjoe: as mentioned above, I think it's probably worth explicitly returning the different error types rather than piggy backing on IOExceptions

ellisjoe (approved):

PR #887, opened 2025-05-12
Do not delete keymaterial before file

mswintermeyer:
## Before this PR
Two bugs exist in `EncryptedFileSystem` that could cause a `KeyMaterial` to be deleted before the file that it encrypts.

First, if `rename` returns `false` but the destination file still exists, we should never delete the destination's `KeyMaterial`. This could happen when trying to rename a file when the destination file already exists, though that means there's likely a bug with the code that called rename. It could also signify an edge case with the `rename` code itself, e.g. `AzureNativeFileSystemStore#rename` can succeed at renaming the file but still not fully succeed because of an error at the end of the file when calling `AzureNativeFileSystemStore#safeDelete` to try to delete the source file.

Second, `delete` should be the inverse of `create`. When we create a file, we create its `KeyMaterial` before the actual file. Thus, on delete, we should flip that order and delete the file before the `KeyMaterial`. Only if the file actually got deleted should we delete its `KeyMaterial`.

## After this PR

==COMMIT_MSG==
Fix some edge cases that could cause a KeyMaterial to be deleted before the file that it encrypts
==COMMIT_MSG==

### DIFF
--- a/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
@@ -145,7 +145,7 @@ public final class EncryptedFileSystem extends DelegatingFileSystem {
 
         if (renamed) {
             tryRemoveKey(src);
-        } else {
+        } else if (!fs.exists(dst)) {
             tryRemoveKey(dst);
         }
 
@@ -172,9 +172,14 @@ public final class EncryptedFileSystem extends DelegatingFileSystem {
             throw new UnsupportedOperationException("EncryptedFileSystem does not support recursive deletes");
         }
 
-        // Interrupted deletes should be resumable. They are expected to be retried.
-        tryRemoveKey(path);
-        return fs.delete(path, false);
+        boolean success = fs.delete(path, false);
+
+        if (success || !fs.exists(path)) {
+            // Interrupted deletes should be resumable. They are expected to be retried.
+            tryRemoveKey(path);
+        }
+
+        return success;
     }
 
     @Override
--- a/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
+++ b/hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
@@ -137,6 +137,11 @@ public final class EncryptedFileSystem extends DelegatingFileSystem {
 
     @Override
     public boolean rename(Path src, Path dst) throws IOException {
+        // Should not overwrite if a file already exists in the destination
+        if (fs.exists(dst)) {
+            return false;
+        }
+
         // Copy key material first so the encrypted file always has key material in the key store even if the
         // put or rename fails
         KeyMaterial keyMaterial = keyStore.get(src.toString());
@@ -146,6 +151,8 @@ public final class EncryptedFileSystem extends DelegatingFileSystem {
         if (renamed) {
             tryRemoveKey(src);
         } else if (!fs.exists(dst)) {
+            // The reason we check again if the destination exists before removing the key, is because there are edge
+            // cases where FileSystem#rename can create the new file but still claim failure.
             tryRemoveKey(dst);
         }
 
--- a/hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
+++ b/hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
@@ -264,6 +264,20 @@ public final class EncryptedFileSystemTest {
         boolean renamed = mockedEfs.rename(path, newPath);
 
         assertThat(renamed).isFalse();
+        verify(mockKeyStore, never()).put(eq(newPath.toString()), any());
+        verify(mockKeyStore, never()).remove(path.toString());
+        verify(mockKeyStore, never()).remove(newPath.toString());
+    }
+
+    @Test
+    public void testRename_failedRenameButDestinationFileExistsAfterCreation() throws IOException {
+        when(mockFs.exists(newPath)).thenReturn(false, true);
+        when(mockFs.rename(path, newPath)).thenReturn(false);
+
+        boolean renamed = mockedEfs.rename(path, newPath);
+
+        assertThat(renamed).isFalse();
+        verify(mockKeyStore).put(eq(newPath.toString()), any());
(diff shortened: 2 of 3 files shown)


estherg77 (dismissed):

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -145,7 +145,7 @@ public boolean rename(Path src, Path dst) throws IOException {
 
         if (renamed) {
             tryRemoveKey(src);
-        } else {
+        } else if (!fs.exists(dst)) {
```
ellisjoe, line 153: I think we need to do this `fs.exists(dst)` check at the top of the method. At this point `keyStore.put(...)` has already been called and would have overwritten the key material.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -172,9 +172,14 @@ public boolean delete(Path path, boolean recursive) throws IOException {
...
         }
 
-        // Interrupted deletes should be resumable. They are expected to be retried.
-        tryRemoveKey(path);
-        return fs.delete(path, false);
+        boolean success = fs.delete(path, false);
+
+        if (success || !fs.exists(path)) {
+            // Interrupted deletes should be resumable. They are expected to be retried.
+            tryRemoveKey(path);
```
ellisjoe, line 186: I think in both `create` and `encrypt` we 1) create the file and then 2) store the key material, so I believe this is doing the inverse operation already by removing the key first and then deleting the file?

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
```diff
@@ -256,6 +256,18 @@ public void testRename_failedRename() throws IOException {
...
+    @Test
+    public void testRename_failedRenameBecauseDestinationFileAlreadyExists() throws IOException {
+        when(mockFs.exists(newPath)).thenReturn(true);
+        when(mockFs.rename(path, newPath)).thenReturn(false);
+
+        boolean renamed = mockedEfs.rename(path, newPath);
+
+        assertThat(renamed).isFalse();
+        verify(mockKeyStore, never()).remove(path.toString());
+        verify(mockKeyStore, never()).remove(newPath.toString());
```
ellisjoe, line 282: I think if you added `verify(mockKeyStore, never()).put(newPath.toString(), any());` you'd end up getting a failure

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: Maybe my question is: what does inverting these calls fix exactly? as soon as either of the file or key material are deleted the file should no longer be readable, so I image there's some other weird error state we're trying to kill?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
mswintermeyer: I can add the check at the top! That said, we'd also want it here unfortunately. As I call out in the PR unfortunately, there are two edge cases for `rename` - one where the file already exists in the destination, and another where rename does properly create the file in the new location, but it still returns false.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
mswintermeyer: Hmm fair, what happens is sort of in the middle here. `create` and `encrypt` create an empty file before they add the key material, but they don't actually return the file and thus add contents to it yet. The key material file would get created before either method returns, and thus before any contents get added to the file.

We've recently seen issues where a file existed with its full content, but without its associated key material file, so I'm trying to remove all ways for that to happen.

hadoop-crypto/src/test/java/com/palantir/crypto2/hadoop/EncryptedFileSystemTest.java
mswintermeyer: Updated tests!

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: Ok so the check at the top deal with the first edge case (file already exists at the destination and we exit immediately).

How can the second edge case happen? Like `fs.rename(src, dst)` actually succeeds but then return `false` even though it was successful?

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: Ok so for the scenario you're describing, do we think what happened was an EFS file was issued a delete, the key material was deleted successfully, and then failed before the file was deleted? If the delete was issued but did not succeed, is it not being retried ever?

I guess I'm not sure how inverting this fixes things? You'll still end up in a state where one of the two files necessary exists while the other may not after a partially executed delete

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
ellisjoe: I guess the other thing is that if a file already exists at dst I feel like we should be relying on the `fs.rename` behavior? Which either should fail because a file already exists, at which point I'd expect `tryRemoveKey(dst)` to be executed, or it actually overwrites the dst file in which case rename success and we run `tryRemoveKey(src)` which seems right?

mswintermeyer: After collecting some more data, I've confirmed that hadoop-azure can run `rename`, return false indicating that it failed, but it leaves around cruft: in addition to it properly leaving the source file around, the destination file can exist too. And that destination file might only be partially written to, or it could even have the same file size as the source. I verified that when we've seen this behavior, the destination file didn't exist before calling `rename` but as mentioned it existed after calling it, even if it fails.

Going to solve this in client code instead for simplicity.

PR #901, opened 2025-11-25
S3 files contain metadata to uniquely associate them with keys

mswintermeyer:
## Before this PR
There's currently a race condition possible that can make files impossible to decrypt, if a file is re-uploaded to s3.

If a file is uploaded to S3 multiple times simultaneously, there are no guarantees on which write wins. Thus, if a service implementing the S3 API takes a long time to process a file creation request (e.g. while ceph is resharding), and the file is re-uploaded during that time, there are no guarantees on which write wins. Worst case: the main file's write from one upload could win, but the key material file's write from the other upload could win. The file is now undecryptable since we do not have the key material it was encrypted with.

## After this PR
Fix the race for s3, by always using a unique suffix/identifier for the file key (so the key material files won't ever overwrite each other) and attaching metadata to the primary file with that identifier.

There's now a risk that if a file is re-uploaded, two key material files could exist for it, and one will be useless and never get deleted. But that's reasonable and small tradeoff, which is much better than having files that cannot be decrypted.

==COMMIT_MSG==
S3 files contain metadata to uniquely associate them with keys, to protect against a race condition when a file is uploaded twice
==COMMIT_MSG==


hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -56,6 +61,9 @@ public final class EncryptedFileSystem extends DelegatingFileSystem {
 
     private static final SafeLogger log = SafeLoggerFactory.get(EncryptedFileSystem.class);
     private static final String DEFAULT_CIPHER_ALGORITHM = AesCtrCipher.ALGORITHM;
+    private static final String S3_METADATA_KEY = "crypto-file-key-suffix";
+    private static final String S3_METADATA_KEY_HEADER =
+            Constants.FS_S3A_CREATE_HEADER + ".x-amz-meta-" + S3_METADATA_KEY;
```
mswintermeyer, line 66: https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata

hadoop-crypto/build.gradle
```diff
@@ -11,6 +11,10 @@ dependencies {
     implementation 'com.palantir.safe-logging:preconditions'
     implementation 'com.palantir.safe-logging:safe-logging'
     implementation 'com.palantir.seek-io:seek-io'
+    implementation ('org.apache.hadoop:hadoop-aws') {
+        exclude group: 'software.amazon.awssdk', module: 'bundle'
+    }
+    implementation 'software.amazon.awssdk:s3'
```
mswintermeyer, line 17: s3 dep is much smaller than the full bundle, and many services that pull in hadoop-aws swap the bundle for explicit deps so that they can pull in transitive versions to fix vulns

hadoop-crypto/build.gradle
```diff
@@ -24,12 +28,8 @@ dependencies {
 
 shadowJar {
     mergeServiceFiles()
-
     dependencies {
-        // The shadow jar is expected to be added to the Hadoop cli classpath
-        // which comes with these deps
-        exclude(dependency('org.apache.hadoop:hadoop-client-api'))
-        exclude(dependency('org.apache.hadoop:hadoop-client-runtime'))
+        include(dependency('com.google.guava:guava'))
```
mswintermeyer, line 32: Now that we're pulling in so many more hadoop deps (hadoop-aws, and all its transitives), and those should be excluded anyways, I figured it's simpler to just define the shadow jar in terms of what it was designed for in the first place: guava

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -117,21 +133,52 @@ public FSDataOutputStream create(
...
 
-        return encrypt(outputStream, path);
+        return encrypt(outputStreamBuilder, path);
     }
 
-    private FSDataOutputStream encrypt(FSDataOutputStream encryptedStream, Path path) throws IOException {
+    private FSDataOutputStream encrypt(FSDataOutputStreamBuilder<?, ?> encryptedStreamBuilder, Path filePath)
+            throws IOException {
+        Optional<String> keyPathSuffix = Optional.empty();
+        if (encryptedStreamBuilder instanceof CreateFileBuilder) {
```
mswintermeyer, line 162: Note: We only go through the codepaths that change the keyPathSuffix if we get a `CreateFileBuilder`, aka it's an s3 compatible store.

I'd definitely be up for also adding a configuration flag, and piping it through (StandaloneEncryptionService#initialize -> EncryptedFileSystem constructor), so that this new behavior is opt-in.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/EncryptedFileSystem.java
```diff
@@ -174,15 +218,44 @@ public boolean delete(Path path, boolean recursive) throws IOException {
         }
 
         // Interrupted deletes should be resumable. They are expected to be retried.
-        tryRemoveKey(path);
+        tryRemoveKey(toKeyPath(path, getKeyPathSuffix(path)));
         return fs.delete(path, false);
     }
 
+    public boolean keyFileExists(Path path) throws IOException {
```
mswintermeyer, line 225: Now that clients can't call `fs#exists(path + FileKeyStorageStrategy.EXTENSION)` (the key path won't always be at the same location), I wanted to introduce some way to expose whether the key path exists. Had to pipe it through StandaloneEncryptedFileSystem as well.

hadoop-crypto/src/main/java/com/palantir/crypto2/hadoop/PathConvertingFileSystem.java
```diff
@@ -93,6 +97,11 @@ public FSDataOutputStream create(
         return delegate.create(to(path), permission, flags, bufferSize, replication, blockSize, progress, checksumOpt);
     }
 
+    @Override
+    public FSDataOutputStreamBuilder<?, ?> createFile(Path path) {
```
mswintermeyer, line 101: This has technically always been missing and a potential bug, if anyone used hadoop-crypto with this method, that since was added to hadoop. It's necessary for my new use case though

mswintermeyer:
Since this adds a decent amount of complexity, going to fork the two classes instead to not affect other consumers

PR #936, opened 2026-09-16
Modernize Gradle plugins

schlosna:
## Before this PR

The repository used Gradle 8.5, older Palantir build plugins, the legacy Shadow plugin, and Java 11 CircleCI images. This prevented the build from adopting the current Gradle, JDK, Baseline, and Error Prone toolchain.

Linux CI also configured native crypto library lookup paths incorrectly. Apache Commons Crypto 1.2.0, the most recent full release, cannot initialize its OpenSSL implementation with some OpenSSL 3 installations, so a test that unconditionally expected Apache-backed streams failed on modern Linux environments even though the JCE fallback remained functional.

## After this PR

- Upgrades the wrapper to Gradle 9.7.1 and modernizes the Palantir build plugins, Gradle scripts, wrapper scripts, and Shadow plugin configuration.
- Runs compilation and CircleCI on JDK 21 while continuing to publish Java 17-compatible library bytecode.
- Corrects the CI lookup paths for `libcrypt` and `libcrypto`.
- Retains the most recent full Commons Crypto release, 1.2.0, instead of depending on an unreleased snapshot.
- Uses the availability check from [#938](https://github.com/palantir/hadoop-crypto/pull/938): the Linux Apache-stream test queries `CryptoStreamFactory` for its memoized OpenSSL availability and skips only the native-specific assertions when OpenSSL cannot be initialized. The test uses valid streams managed with try-with-resources; the rest of `CryptoStreamFactoryTest` continues to run against the JCE fallback.
- Configures Mockito as an explicit Java agent for JDK 21, makes annotation-processing dependencies explicit, and enables implicit dependency checks.
- Applies the Java and dependency cleanups required by newer Baseline/Error Prone, including specific safe exception types and current Hadoop APIs.
- Updates JUnit, ASM, and related locked dependencies.

## Possible downsides?

- Building the project now requires a JDK 21-capable environment, although published artifacts remain compatible with Java 17.
- Environments where Commons Crypto 1.2.0 cannot initialize OpenSSL do not run the Apache-specific stream assertions. Native OpenSSL 3 compatibility will need to be re-enabled when it is available in a full Commons Crypto release; JCE behavior remains covered.
- Some failure paths now throw more specific safe runtime exception types, which may be observable to callers that match exact exception classes.


crypto-core/src/test/java/com/palantir/crypto2/io/CryptoStreamFactoryTest.java
```diff
@@ -52,6 +53,7 @@ public void before() {
     @Test
     @EnabledOnOs(OS.LINUX)
     public void ensureDefaultIsApache() {
+        assumeTrue(CryptoStreamFactory.isOpenSslAvailable());
```
schlosna:
```suggestion
        assertThat(CryptoStreamFactory.isOpenSslAvailable()).as("OpenSSL available").isTrue();
```

crypto-core/src/test/java/com/palantir/crypto2/io/CryptoStreamFactoryTest.java
```diff
@@ -17,6 +17,7 @@
 package com.palantir.crypto2.io;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
```
schlosna:
```suggestion
```

crypto-core/src/test/java/com/palantir/crypto2/io/CryptoStreamFactoryTest.java
schlosna: https://github.com/palantir/hadoop-crypto/pull/938

build.gradle
```diff
@@ -45,12 +48,18 @@ subprojects {
     apply plugin: 'com.palantir.external-publish-jar'
 
     tasks.check.dependsOn javadoc, checkImplicitDependencies, checkUnusedDependencies
+
+    dependencies {
+        testRuntimeOnly 'org.mockito:mockito-subclass'
+    }
```
pkoenig10: We shouldn't need this - the default inline mock maker should be sufficient.

Plus we shouldn't be blanket adding dependencies like this.

build.gradle
```diff
@@ -45,12 +48,18 @@ subprojects {
...
 }
 
 javaVersions {
+    javaCompiler = 21
```
pkoenig10:
```suggestion
    javaCompiler = 25
```

build.gradle
```diff
@@ -45,12 +48,18 @@ subprojects {
...
     libraryTarget = 17
     runtime = 21
 }
+
 jdks {
-    daemonTarget = "21"
+    daemonTarget = 21
```
pkoenig10:
```suggestion
    daemonTarget = 25
```

build.gradle
schlosna: removed

build.gradle
schlosna: let's do JDK bumps in follow on PR, I'd like excavator to handle bumping JDK & gradle 9

pkoenig10 (dismissed):

build.gradle
schlosna: bumped JDK 25 [42694d3](https://github.com/palantir/hadoop-crypto/pull/936/commits/42694d3d1ff4b0ebcb68a2b08555549ae84834cd)

pkoenig10 (approved):

PR #937, opened 2026-09-16
Add explicit immutables dependency

schlosna:
## Before this PR

`KeyMaterial` declared `@Value.Style` directly, including `Value.Style.ImplementationVisibility.PACKAGE`. Because `KeyMaterial` is public, that enum type was recorded in its classfile and effectively exposed Immutables as an implicit runtime dependency to library consumers. The Immutables annotations artifact was not declared explicitly on the `crypto-keys` compile classpath, so consumers could encounter missing Immutables types despite those types not being part of the intended API.

This is the same class of issue addressed by [conjure-java-runtime-api#1689](https://github.com/palantir/conjure-java-runtime-api/pull/1689).

## After this PR

`crypto-keys` explicitly declares `org.immutables:value::annotations` as a `compileOnly` dependency.

The Immutables configuration is encapsulated in a source-retained, package-level `ImmutablesStyle` annotation. `KeyMaterial.class` therefore no longer references `Value.Style` or `Value.Style.ImplementationVisibility`, so consumers are not required to provide Immutables at runtime.

`checkImplicitDependencies` is also included in `check` for every subproject to catch future undeclared compile dependencies.

## Possible downsides?

The package-level style applies to any future Immutables value type added to `com.palantir.crypto2.keys`. This is intentional and keeps those implementations package-private without adding a runtime dependency for consumers.


pkoenig10 (approved):
While we're here let's migrate off the `org.inferred processors` plugin to the modern `annotationProcessor` configuration.

schlosna:
> migrate off the `org.inferred processors` plugin to the modern `annotationProcessor` configuration

https://github.com/palantir/hadoop-crypto/pull/939 to `Migrate off the gradle-processors plugin`

PR #938, opened 2026-09-16
Only test Apache streams when OpenSSL is available

schlosna:
## Before this PR

`CryptoStreamFactoryTest` assumed every Linux environment had a compatible native OpenSSL implementation, causing the Apache stream assertion to fail when Commons Crypto could not initialize OpenSSL.

For example, https://github.com/palantir/hadoop-crypto/pull/930 fails with newer ubuntu due to OpenSSL 3 upgrade as OpenSSL 1.x is past end-of-life, for which latest Apache commons-crypto release 1.2.0 does not support OpenSSL 3; however, unreleased commons-crypto 1.2.1-SNAPSHOT does support OpenSSL 3.

## After this PR

The test queries CryptoStreamFactory's memoized OpenSSL availability and uses a JUnit assumption before making Apache-specific stream assertions.

Verified with Java 21:

    ./gradlew :crypto-core:spotlessApply :crypto-core:test --tests com.palantir.crypto2.io.CryptoStreamFactoryTest

## Possible downsides?

The Apache-specific type assertions are skipped on hosts without a compatible OpenSSL implementation; the remainder of CryptoStreamFactoryTest still runs.


ellisjoe:
The only thing I'm worried about here is that once we take the Ubuntu upgrade, we won't be testing the Apache crypto path anymore. I think it's a reasonable goal to remove Apache crypto altogether once we verify the performance of the JCE based cipher's is equivalent, so this is maybe fine as long as we actually do the work to remove Apache commons crypto altogether.

schlosna:
closing in favor of keeping 3.x requiring OpenSSL testing and https://github.com/palantir/hadoop-crypto/pull/945 removing OpenSSL & Apache commons-crypto in 4.x

PR #939, opened 2026-09-16
Migrate off the gradle-processors plugin

schlosna:
## Before this PR

Every subproject applied the legacy `org.inferred.processors` plugin and received a global Immutables `processor` dependency, even though only `crypto-keys` uses the Immutables annotation processor. This relied on a custom, deprecated processor configuration instead of Gradle's standard annotation-processing configurations.

## After this PR

The `org.inferred.processors` plugin, its buildscript classpath, and the global `processor` dependency are removed.

`crypto-keys` now declares `annotationProcessor 'org.immutables:value'` directly alongside its compile-only Immutables annotations dependency. JMH continues to use the existing `testAnnotationProcessor` and `testCompileOnly` declarations.

Validated with `:crypto-keys:check` and the implicit/unused dependency checks across all modules.

## Possible downsides?

Future modules that introduce annotation-processed sources must declare their own `annotationProcessor` dependency instead of inheriting one globally. This makes processor usage explicit and scoped to the modules that require it.


pkoenig10 (approved):
Thanks!

PR #944, opened 2026-09-16
Move Commons Crypto to optional apache-crypto module

schlosna:
## Before this PR

`crypto-core` exposes Apache Commons Crypto as an API dependency and selects its OpenSSL implementation for AES/CTR when available. This introduces a native runtime dependency and makes consumers vulnerable to Commons Crypto and OpenSSL compatibility failures even when they only need the JDK implementation.

## After this PR

- `crypto-core` and `hadoop-crypto` rely solely on JDK-provided crypto providers.
- `ApacheCiphers` and `ApacheCtrDecryptingSeekableInput` retain their existing packages and public APIs in the optional `com.palantir.hadoop-crypto2:apache-crypto` artifact.
- Commons Crypto and JNA resolve only for consumers that select `apache-crypto`.
- Apache-specific benchmark coverage and obsolete OpenSSL fallback documentation are removed from `crypto-core`.

## Possible downsides?

Consumers that directly use the Apache-specific classes must add the new `apache-crypto` artifact. `CryptoStreamFactory` now always uses the JDK provider, which changes performance characteristics but preserves the AES ciphertext format.

See benchmarks comparing JDK crypto provider versus OpenSSL 3 via Apache commons-crypto in https://github.com/palantir/hadoop-crypto/pull/943 & https://github.com/palantir/hadoop-crypto/blob/davids/benchmark/benchmark-results/encryption/README.md


ellisjoe:
I kinda want to just fully remove commons crypto which has been annoying for awhile now. An alternative proposal is here: https://github.com/palantir/hadoop-crypto/pull/945

Let me know if you think there's a specific reason you think we should keep it around though. I'm thinking we release 4.0 either way so we can slow roll the upgrade internally and make sure we don't see any unexpected performance regressions.

schlosna:
closing in favor of https://github.com/palantir/hadoop-crypto/pull/945

Issue #105, opened 2018-01-23
Override all create and open methods in EncryptedFileSystem


ellisjoe:
Now that EncryptedFileSystem extends FilterFileSystem, many of the create and open methods will not get forwarded to the create and open methods currently implemented by FilterFileSystem (see: https://github.com/palantir/hadoop-crypto/pull/104). To fix this we either need to go back to implementing FileSystem and overriding all of the methods that are known to be broken (eg: copyFromLocal), or continue to extend FilterFilesystem and override all of the create and open methods.

PR #963, opened 2026-09-17
Override all create and open methods in EncryptedFileSystem

ZayanKhan-12:
## Before this PR

`EncryptedFileSystem` extends `DelegatingFileSystem`, which extends `FilterFileSystem`. As #105 notes, `FilterFileSystem` forwards a number of create/open entry points **straight to the wrapped delegate**, so overriding `create`/`open` is not enough — the undecorated delegate stream still escapes through the others.

For `EncryptedFileSystem` this is silent and severe. Against `develop` today:

```java
try (FSDataOutputStream os = efs.createFile(file).build()) {
    os.write(data);                       // written as PLAINTEXT, no key material stored
}
ByteStreams.toByteArray(efs.openFile(path).build().get());
                                          // returns raw CIPHERTEXT
```

The tests added here fail on `develop` with exactly that:

```
testCreateFile_encryptsAndStoresKeyMaterial:
  Expecting actual not to be null            <- keyStore.get(path) == null

testOpenFile_decrypts:
  expected: [115, 111, 109, 101, 32, ...]    <- "some data to encrypt"
   but was: [-48, -72, 29, -33, 29, ...]     <- ciphertext
```

The full set of leaking methods in `FilterFileSystem` (Hadoop 3.3.4): `createFile`, `appendFile`, `openFile(Path)`, `openFile(PathHandle)`, `openFileWithOptions` (both overloads), `createNonRecursive`, `primitiveCreate`, and `open(PathHandle, int)`.

Note that `createFile(path).build()` routes through `createNonRecursive`, not `create` — `FSDataOutputStreamBuilder.recursive` defaults to `false`.

## After this PR

`DelegatingFileSystem` now makes good on its contract: **every** way of creating or opening a file funnels through this class's `create` / `open` / `append`. That is the whole fix — both existing subclasses inherit it, and any future one does too.

- `createFile`, `appendFile`, `openFile(Path)`, `openFile(PathHandle)` return builders bound to `this` (via Hadoop's `createDataOutputStreamBuilder` / `createDataInputStreamBuilder` hooks) instead of the delegate's builders.
- `openFileWithOptions` (both overloads) resolves to `this.open(...)`, mirroring `FileSystem`'s base implementation including `rejectUnknownMandatoryKeys`. This is required: it is what the `openFile` builder calls from `build()`.
- `createNonRecursive` and `primitiveCreate` are implemented in terms of `create`/`append`.
- `open(PathHandle, int)` and `createPathHandle` now throw instead of forwarding. A `PathHandle` is an opaque, delegate-specific reference that a decorator cannot translate back into a path, and `EncryptedFileSystem` needs the path to look up key material. Refusing to mint handles as well means callers fail fast rather than at read time.

These re-routing methods are `final`, so a subclass cannot accidentally re-open the hole; `create`/`open`/`append` remain the extension points.

`PathConvertingFileSystem.createFile` was a local patch for one instance of this bug and is now redundant — deleted.

Tests: 13 added, covering each entry point both generically (`DelegatingFileSystemTest`, recording subclass) and end-to-end through real encryption (`EncryptedFileSystemTest`, asserting the bytes on the delegate really are ciphertext). 10 of the 13 fail without the main-source change; the 11th, `testPrimitiveCreateRoutesThroughCreate`, does not even compile against `develop`, since `primitiveCreate` only becomes visible to the test package once `DelegatingFileSystem` re-declares it. `./gradlew build` passes.

==COMMIT_MSG==
`EncryptedFileSystem` no longer writes plaintext or returns ciphertext through the create/open entry points that `FilterFileSystem` forwards directly to the underlying FileSystem (`createFile`, `appendFile`, `openFile`, `createNonRecursive`, `primitiveCreate`). Reads and writes via `PathHandle` are now rejected rather than silently bypassing encryption.
==COMMIT_MSG==

## Possible downsides?

Three behaviour changes, each replacing a silent correctness bug with a loud failure:

1. **`PathHandle` operations now throw.** Previously they returned ciphertext from `EncryptedFileSystem` — unusable data, not a working feature.
2. **`createNonRecursive` is no longer atomic.** It checks the parent up front and then calls `create`, so there is a benign race where a concurrently created parent turns a would-be failure into a success. The alternative is writing plaintext, and several Hadoop filesystems (e.g. S3A) do not offer atomic non-recursive create either.
3. **`createFile`/`appendFile`/`openFile` now honour `EncryptedFileSystem`'s restrictions** — most visibly, `appendFile(...).build()` throws "appending to encrypted files is not supported", matching `append`, where before it appended plaintext to a ciphertext file.

Files written through the affected paths before this change are still plaintext with no key material, and remain unreadable through `EncryptedFileSystem`; this PR does not attempt to detect or migrate them.

Out of scope, flagged for a follow-up: `concat` and `truncate` are forwarded to the delegate by `FilterFileSystem` too, and are also wrong for encrypted files (each file has its own key and IV). They are not create/open methods, so I left them out of this change rather than widen it — happy to open a separate issue.

**The second commit adds a `CLAUDE.md`** documenting the build and the decoration invariant above, for anyone using Claude Code in this repo. It is independent of the fix — drop that commit if it is not wanted here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


palantirtech:
Thanks for your interest in [palantir/hadoop-crypto](https://github.com/palantir/hadoop-crypto), @ZayanKhan-12! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #964, opened 2026-09-17
Add support for recursive delete in EncryptedFileSystem

ZayanKhan-12:
## Before this PR

`EncryptedFileSystem.delete(path, true)` throws `UnsupportedOperationException`. From #117, the blocker was never the traversal — it was that `KeyStorageStrategy` had no way to participate:

> We need to make `KeyStorageStrategy` expose recursive delete and rename options in order for `EncryptedFileSystem` to support recursive delete and move. — @ellisjoe

The documented workarounds are to use `StandaloneEncryptedFileSystem`, or to delete on the underlying FileSystem and leave `EncryptedFileSystem` out of it. The latter silently strands key material for every deleted file when the key store is not `FileKeyStorageStrategy`.

This takes the approach that comment describes, rather than the one in #118 (closed), which recursed a level at a time on the client, called `keyStore.remove` for directories as well as files, and used the throwing `remove` — so a single already-removed key aborted the whole delete partway through.

## After this PR

**`KeyStorageStrategy` gains a batch removal** — a `default` method, so every existing implementation keeps compiling and working untouched:

```java
default void remove(Set<String> fileKeys)
```

It attempts *every* file key even when some fail, then throws with the individual failures attached as suppressed exceptions. That property is the point: with a plain `fileKeys.forEach(this::remove)`, one unremovable key strands the key material of every file after it. Stores backed by a remote service can override it to issue one batched call; `ChainedKeyStorageStrategy` overrides it to hand the whole batch to each strategy rather than degrading to one call per key per strategy.

**`EncryptedFileSystem.delete` implements the recursive case** by enumerating the tree, removing the key material, then letting the delegate do the data delete in its own single recursive call.

Two details worth a reviewer's eye:

- **Child paths are resolved against the caller's path**, not read off the `FileStatus`. `create` keys key material by `path.toString()` as the caller spelled it, while `FileStatus.getPath()` is fully qualified (`file:/tmp/x` vs `/tmp/x`). Using the `FileStatus` path directly looks correct and silently removes nothing.
- **Traversal uses `getFileStatus`/`listStatus`, not `listFiles`.** `PathConvertingFileSystem` overrides the former but not the latter, so `listFiles` would return unconverted paths from the raw delegate.

Key material is removed before the data, matching the existing non-recursive path, so an interrupted delete stays resumable: the files are still there, so the remaining keys can be enumerated and removed again on retry. Removal is best effort — failure is logged and the data is deleted regardless, which is the existing `tryRemoveKey` behaviour.

Tests: 9 added. The 4 `EncryptedFileSystem` delete tests all fail against `develop`. One covers the `EncryptedFileSystem` + `FileKeyStorageStrategy` combination specifically, since there the recursive listing also turns up the `.keymaterial` files — removing "key material for key material" is a no-op and the key files are deleted with the rest of the tree, but that is worth pinning down rather than asserting. `./gradlew build` passes.

==COMMIT_MSG==
`EncryptedFileSystem` now supports recursive deletes, removing the `KeyMaterial` for every file in the tree instead of throwing `UnsupportedOperationException`. `KeyStorageStrategy` has a new `remove(Set<String>)` default method which implementations backed by a remote store can override to remove key material in a single batched call.
==COMMIT_MSG==

## Possible downsides?

- **`delete(path, true)` no longer throws.** Anyone relying on that as a guard against accidental recursive deletes loses it. This is the requested behaviour change, and `StandaloneEncryptedFileSystem` already deletes recursively today.
- **Cost is O(files) in key-store calls.** Enumerating a large tree costs a `listStatus` per directory, and key removal costs one call per file unless the strategy overrides the batch method. That is inherent to key material being per file; the batch hook exists so stores that can amortise it do.
- **Not atomic.** Key material is removed first, so a delete interrupted midway leaves files whose key material is already gone — they are unreadable, but they were about to be deleted. Retrying completes the delete. This matches the existing non-recursive behaviour rather than introducing it.
- Key material orphaned by *previous* deletes done directly on the underlying FileSystem is not cleaned up; this PR does not attempt to detect it.

Out of scope: recursive `rename`, also mentioned in #117. It needs a copy-and-rekey or a key-store move primitive, which is a larger design question than the delete, so I kept this PR to what the issue title asks for. `AsyncKeyStorageStrategy` is untouched for the same reason — `EncryptedFileSystem` does not use it.

Note: this branch is cut from `develop` and is independent of my #963. Both touch `EncryptedFileSystemTest`, so whichever lands second will need a trivial conflict resolution. The `CLAUDE.md` you may be expecting alongside this is in #963 — I deliberately did not duplicate it here, since that would guarantee a conflict.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


palantirtech:
Thanks for your interest in [palantir/hadoop-crypto](https://github.com/palantir/hadoop-crypto), @ZayanKhan-12! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

